PR #39, opened 2025-04-07
Introduced modules

MoSchmidt:
4 modules (one for each dev) were created. Each module also contains the same packages (api, service, repository & dto) which contain empty classes, that can be removed.


ani-ib (approved):
new structure looks great, I ran it locally and it works well

PR #63, opened 2025-04-17
`Development`: Added Custom Fields data structure

MoSchmidt:
#### General

#### Server

### Motivation and Context

Custom Fields allow professor to define custom questions they want to include in their Job postings.

### Description

- Added CustomField class that defines the structure and answer possibilities of such field
- Added CustomFieldAnswer class which stores the actual answer (also documents) specific to one application

### Steps for Testing

Verify if the relations between the entities are set correctly


src/main/java/de/tum/cit/aet/job/domain/CustomField.java
```diff
@@ -0,0 +1,43 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    @Column(name = "custom_field_id", nullable = false)
+    private UUID customFieldId;
+
+    @ManyToOne
+    @JoinColumn(name = "job_id")
+    private Job job;
+
+    private int sequence;
```
ani-ib, line 28: What is the use of the "sequence" column?

src/main/java/de/tum/cit/aet/job/domain/CustomField.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private CustomFieldType type;
```
ani-ib, line 30:
```suggestion
    @Column(name = "custom_field_type")
    private CustomFieldType type;
```
Suggested adding @Column(name = "custom_field_type") to match the database’s snake case naming convention

ani-ib (changes requested):
Code looks good overall, I added a suggestion and a question about the CustomField Entity's "sequence" column that wasn't visible in the [Database Structure](https://confluence.aet.cit.tum.de/spaces/AP/pages/258573136/Database+Structure ) page on confluence

src/main/java/de/tum/cit/aet/application/domain/CustomFieldAnswer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @ManyToOne
+    @JoinColumn(name = "custom_field_id")
+    private CustomField customField;
+
+    @ManyToOne
+    @JoinColumn(name = "application_id")
+    private Application application;
+
+    @Convert(converter = StringListConverter.class)
```
ani-ib, line 32: Didn't we want to avoid JSON as it restricts us? Or do we intend to still save some fields as JSON?

src/main/java/de/tum/cit/aet/job/domain/CustomField.java
MoSchmidt: The sequence field will ensure that the custom fields are always displayed in the correct order in which they have been set by the professor. This is important if questions have to be asked in a specific order

src/main/java/de/tum/cit/aet/application/domain/CustomFieldAnswer.java
MoSchmidt: I would store those fields as serialized JSON inside a column, as we won't have too many answers per question, and we are not editing them at the same time. This will keep the data structure smaller and we wont have to add extra tables

src/main/java/de/tum/cit/aet/job/domain/CustomField.java
MoSchmidt: Thanks, I will fix this

src/main/java/de/tum/cit/aet/application/domain/CustomFieldAnswer.java
az108: Please change this to a Set

src/main/java/de/tum/cit/aet/job/domain/CustomField.java
ani-ib: Alright makes sense, thanks for the clarification

az108: I dont quite get the use of this. Why cant we just store it in job in the correct order. Wont we have a List<CustomField> there?

ani-ib (approved):
Code LGTM - relations between the entities seem to be set correctly

src/main/java/de/tum/cit/aet/job/domain/CustomField.java
MoSchmidt: Yes, we will use a List, but JPA does not maintain the same order of CustomFields when storing it in the database. So when we would load the List of CustomFields again from the database, they could be in a different order. But I assume we always want them to be shown in the same order to create a form where the applicant can enter their information. Thas why we need the sequence field that defines the order. By adding "ORDER BY sequence" we can assure that the CustomFields are always shown in the correct order

az108: alright 👍 thanks for the explanation

PR #64, opened 2025-04-18
`Development`: Adjusted reject reason

Douglas-Sarah:
#### General

#### Server

### Motivation and Context

The client will ask the server for a list of possible reject reasons which can be fetched from this new enum `RejectReason`.
The supervisor can use and adapt those to their needs in the client.
This method is chosen because later on instead of an enum a database can be used and each chair can implement custom Reject Reason Texts.

### Description

- new Enum RejectReason
- removed Reject Reason from ApplicationReview

### Steps for Testing

Prerequisites:

Test if server builds

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1


src/main/java/de/tum/cit/aet/application/domain/ApplicationReview.java
```diff
@@ -30,9 +30,6 @@ public class ApplicationReview {
     @JoinColumn(name = "application_id")
     private Application application;
 
-    @Column(name = "reject_reason")
```
ani-ib, line 33: Could you also add a new attribute for rejectReason in the relevant entity? Right now, the enum is defined, but the field that should use it hasn’t been added yet.

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
```diff
@@ -0,0 +1,25 @@
...
+ * The supervisor can use and adapt those to their needs.
+ * This method is chosen because lateron instead of an enum a database can be used
+ * and each chair can implement custom Reject Reason Texts.
+ */
+
+@Getter
+@AllArgsConstructor
+public enum RejectReason {
+    FAILED_APPLICANT_REQUIREMENTS("FAILED_APPLICANT_REQUIREMENTS"),
+    POSITION_ALREADY_FILLED("POSITION_ALREADY_FILLED"),
```
ani-ib, line 17: Just to clarify: Is it logical to have a rejection reason stating that the position has already been filled, or should the rejection provide a more specific explanation tailored to the applicant’s case?

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
```diff
@@ -0,0 +1,25 @@
...
+    NOT_SUFFICIENT_GRADES("NOT_SUFFICIENT_GRADES");
+
```
ani-ib, line 19:
```suggestion
    STRONGER_CANDIDATES_SELECTED("STRONGER_CANDIDATES_SELECTED"),

```
Another possible suggestion for a rejection reason

ani-ib, line 19:
```suggestion
    LACK_OF_RESEARCH_EXPERIENCE(" LACK_OF_RESEARCH_EXPERIENCE"),
```
Another suggestion for a rejection reason that emphasizes that the applicant does not have enough experience (the name of the enum constant can be refined if needed).

src/main/java/de/tum/cit/aet/application/domain/ApplicationReview.java
Douglas-Sarah: So the idea is that we give the profs different rejection-reason-templates within the client, and they can adjust those templates to the specific cases if they want to. The adjusted reason is then saved in the field "reason". So we technically don't have to save the enum in the database.
The enum is also thought to be a precursor to a database where the different chairs can save their chair-specific rejection reasons. Each rejection reason should provide the template texts for the different languages, but I thought we could reuse the Mail-Template Engine that Moritz might build for the Mail-and-Notification System, so I hinted at it with the function "getStandardText" in the enum Reject_Reason.

Do you think we should save the enum in the database as well?

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
Douglas-Sarah: I would suggest providing the profs with a unspecific standard text, which the profs can tailor if they want to. If they get a very large amount of applications, they may not want to tailor the rejections too much.

Douglas-Sarah:
@ani-ib, thank you for your suggestions, they were very good

src/main/java/de/tum/cit/aet/application/domain/ApplicationReview.java
ani-ib: No you're right, this makes a lot of sense. Thanks for the detailed explanation. Saving the enum does not seem necessary for now.

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+@Getter
+@AllArgsConstructor
+public enum RejectReason {
+    FAILED_APPLICANT_REQUIREMENTS("FAILED_APPLICANT_REQUIREMENTS"),
+    POSITION_ALREADY_FILLED("POSITION_ALREADY_FILLED"),
+    NOT_SUFFICIENT_GRADES("NOT_SUFFICIENT_GRADES"),
+    STRONGER_CANDIDATES_SELECTED("STRONGER_CANDIDATES_SELECTED"),
+    LACK_OF_RESEARCH_EXPERIENCE(" LACK_OF_RESEARCH_EXPERIENCE");
```
ani-ib, line 20:
```suggestion
    LACK_OF_RESEARCH_EXPERIENCE("LACK_OF_RESEARCH_EXPERIENCE");
```
I added an extra space by mistake to my previous suggestion 😅

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
ani-ib: Alright, makes sense. Thanks for clarifying

src/main/java/de/tum/cit/aet/application/domain/RejectReason.java
Douglas-Sarah: Good correction- I didn't notice 😅

ani-ib (approved):
Code LGTM - tested and server builds

az108 (approved):
Code LGTM 👍

PR #70, opened 2025-04-23
`Development`: Adjust classes to use auditing superclass

MoSchmidt:
#### General

#### Server

### Motivation and Context

We need the Auditing entries **createdAt** and **lastModifiedAt**for many entities. Defining them in a MappedSuperClass allows us to extend this class when the auditing fields are required, without defining those fields manually in each entity.

I also adapted the User and the Applicant class to implement the "Joined Table Strategy". This allows the Applicant entity to extend the User and access all its methods and attributes directly from the applicant. Also the audit entries can now directly be accessed from the applicant.

### Description

- Adapted the AbstractAuditingEntity and made it simpler
- Adapted existing entities to extend this class
- added Inheritance strategy to User class and make Applicant extend User

### Steps for Testing

- verify if all necessary entities extend AbstractAuditingEntity
- check if User Applicant inheritance is set propperly

### DIFF
--- a/src/main/java/de/tum/cit/aet/application/domain/Application.java
+++ b/src/main/java/de/tum/cit/aet/application/domain/Application.java
@@ -1,6 +1,7 @@
 package de.tum.cit.aet.application.domain;
 
 import de.tum.cit.aet.application.constants.ApplicationStatus;
+import de.tum.cit.aet.core.domain.AbstractAuditingEntity;
 import de.tum.cit.aet.evaluation.domain.InternalComment;
 import de.tum.cit.aet.job.domain.Job;
 import de.tum.cit.aet.usermanagement.domain.Applicant;
@@ -10,13 +11,12 @@ import java.util.Set;
 import java.util.UUID;
 import lombok.Getter;
 import lombok.Setter;
-import org.hibernate.annotations.CreationTimestamp;
 
 @Entity
 @Getter
 @Setter
 @Table(name = "applications")
-public class Application {
+public class Application extends AbstractAuditingEntity {
 
     @Id
     @GeneratedValue(strategy = GenerationType.UUID)
@@ -49,8 +49,4 @@ public class Application {
 
     @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "application")
     private Set<InternalComment> internalComments;
-
-    @CreationTimestamp
-    @Column(name = "created_at")
-    private Instant createdAt;
 }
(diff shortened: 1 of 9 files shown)


Douglas-Sarah:
Just for clarification, because it looks good this far:
`CustomField` and `CustomAnswer` don't inherit from the new `AbstractAuditingEntity` because they depend on the `Application`, so the separate `created_at` and `last_modified_at` are irrelevant here, right?
And `ApplicationReview` will only be created once, and probably not updated, so it only needs the `reviewed_at` timestamp and not `last_modified_at`.
Why does `CustomField` not inherit from `AbstractAuditingEntity`? I think both timestamps would be useful here.

ani-ib (approved):
Code LGTM – Introducing a common auditing superclass improves consistency across entities and supports maintainability.

MoSchmidt:
> Just for clarification, because it looks good this far: `CustomField` and `CustomAnswer` don't inherit from the new `AbstractAuditingEntity` because they depend on the `Application`, so the separate `created_at` and `last_modified_at` are irrelevant here, right? And `ApplicationReview` will only be created once, and probably not updated, so it only needs the `reviewed_at` timestamp and not `last_modified_at`. Why does `CustomField` not inherit from `AbstractAuditingEntity`? I think both timestamps would be useful here.

Yes, that was also my idea for your first two questions.
I thought the same for CustomField as this gets defined together with a Job, and we won't allow editing this field after it was created (only when no application was posted for this joby yet, you are still able to edit). I agree, for this it would be a valid case to extend the AbstractAuditingEntry. What do you think?

Douglas-Sarah:
> > Just for clarification, because it looks good this far: `CustomField` and `CustomAnswer` don't inherit from the new `AbstractAuditingEntity` because they depend on the `Application`, so the separate `created_at` and `last_modified_at` are irrelevant here, right? And `ApplicationReview` will only be created once, and probably not updated, so it only needs the `reviewed_at` timestamp and not `last_modified_at`. Why does `CustomField` not inherit from `AbstractAuditingEntity`? I think both timestamps would be useful here.
> 
> Yes, that was also my idea for your first two questions. I thought the same for CustomField as this gets defined together with a Job, and we won't allow editing this field after it was created (only when no application was posted for this joby yet, you are still able to edit). I agree, for this it would be a valid case to extend the AbstractAuditingEntry. What do you think?

Thanks for the clarification, that makes sense. LGTM too

Douglas-Sarah (approved):
Like I said, LGTM too

PR #75, opened 2025-04-23
`Development`: Adjust job, applicant, and application entity fields

ani-ib:
#### General

#### Server

### Motivation and Context

Adjusted the `Job`, `Applicant`, and `Application` entities to align with the latest database schema. This change ensures entity consistency, supports proper relationships (including one-to-one mappings to `Document`), and prepares the server for future business logic and form processing.

### Description

- Updated the `Application` entity to reference `Document` for fields like `cvFile`, `bachelorCertificate`, etc.
- Verified that UUID-based primary and foreign keys match schema constraints.
- Added Draft state for jobs, so drafts can be filtered out in the Job Listing View.

#### Code Review

- [ ] Code Review 1


src/main/java/de/tum/cit/aet/application/domain/Application.java
```diff
@@ -44,6 +45,31 @@ public class Application extends AbstractAuditingEntity {
     @Column(name = "desired_start_date")
     private Instant desiredStartDate;
 
+    @OneToOne
+    @JoinColumn(name = "cv_file_id")
+    private Document cvFile;
+
+    @OneToOne
+    @JoinColumn(name = "reference_file_id")
+    private Document referenceFile;
```
Douglas-Sarah, line 54: Most research jobs I saw demand multiple references - usually at least two. 
Is the single reference file by intentional and was already talked about?
If not:
I would not store the reference in the `Document`, so I suggest a join table. We would need to adjust the database scheme for that as well.
```
@OneToMany(cascade = CascadeType.ALL, orphanRemoval = true)
@JoinTable(
    name = "reference_file_join_documents_application",
    joinColumns = @JoinColumn(name = "application_id"),
    inverseJoinColumns = @JoinColumn(name = "document_id")
)
private Set<Document> referenceFiles;
```

We could also create a new ReferenceFile entity, but that would only make sense if there are any more information to store in that Object, and I think we decided against that.

Same for Applicant.
I can commit this change if we agree on it.

src/main/java/de/tum/cit/aet/usermanagement/domain/Applicant.java
```diff
@@ -62,6 +72,7 @@ public class Applicant extends User {
     @Column(name = "interests")
     private String interests;
 
+    // Is this necessary, now that we are mapping each document separately as a column?
     @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "applicant")
     private Set<Document> documents;
```
Douglas-Sarah, line 77: I would remove that as well - but we need to update the database scheme and AOM for that

src/main/java/de/tum/cit/aet/usermanagement/domain/Applicant.java
Douglas-Sarah, line 1: In the AOM, there are fields that are missing in the database scheme but still seem relevant:

- phoneNumber
- country
- bachelorStartDate
- bachelorEndDate
- masterStartDate
- website
- linkedInUrl
- isTumStudent

Douglas-Sarah (commented):
I note down change suggestions, but I would wait with an implementation until our review meeting today.
A lot of those would require a change in the database scheme / AOM scheme, so I think we would need approval for those.

src/main/java/de/tum/cit/aet/application/domain/Application.java
Douglas-Sarah: After todays meeting we decided to keep it simple and keep the one reference. Any further references will be stored in Customfields, if the professor wants/needs them

src/main/java/de/tum/cit/aet/usermanagement/domain/Applicant.java
ani-ib: The fields linkedInUrl and website will be stored in the User Entity

ani-ib: The bachelorStartDate, bachelorEndDate, masterStartDate, (and masterEndDate) need to be discussed in tumapply channel before finalizing.

ani-ib: Will update Applicant table with tum_student field

src/main/java/de/tum/cit/aet/usermanagement/domain/Applicant.java
Douglas-Sarah: Removed with [this commit](https://github.com/ls1intum/tum-apply/pull/75/commits/d5e8247793d6aa8e5c3270fd1acebda2f9e52878) - removed the Applicant as well from the Document, as we only have OneToOne relations as of now, and not only between Documents and Applicants, so the Applicant in the Document Entity is not necessary

Douglas-Sarah:
Adjusted according to a suggestion from @MoSchmidt : 
- moved rating from ApplicationReview to Application
- changed rating type from int to Integer

marc-fett (approved):

az108 (approved):
Code LGTM 👍

PR #290, opened 2025-07-14
`Development`: Create applicant fix

Douglas-Sarah:
#### General

#### Server

#### Client

### Motivation and Context

This PR handles the issue where users logging in via external providers are unable to create an application.

**Root Cause**: This issue stems from how we currently model User and Applicant entities. On first login, users are created as User and persisted in the UserRepository. When a user attempts to create an application, the system checks for a corresponding Applicant entry in the ApplicantRepository. If none is found, we intended to create one dynamically.
However, since Applicant currently inherits from User (`Applicant extends User` with `@Inheritance(strategy = InheritanceType.JOINED)`), and both share the same primary key, JPA does not allow creating a new Applicant for an existing User. This makes our dynamic conversion from User to Applicant problematic.
### Description

We suggest refactoring the model to use composition instead of inheritance. For example:
This change would give us more flexibility: a User could dynamically gain roles like Applicant, Professor, ResearchGroupAssistant, etc., by associating with different role-specific data objects.
Impact:
We'd need to adapt parts of the backend that directly access Applicant data — but these changes would be caught at compile time, minimizing the risk.
No changes to the actual database schema are required, only the entity mappings and access logic.

### Steps for Testing

- Test whether you can login via email or an external provider
- Test whether you can create an Application

- Test Job System
- Test Evaluation System
- Test Application system
    - [ ] fix error when trying to send application (automatic save does work) 

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1


src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
```diff
@@ -44,25 +52,34 @@ public class ApplicationService {
...
             throw new OperationNotAllowedException("Applicant has already applied for this position");
         }
-        Applicant applicant = applicantRepository
-            .findById(applicantId)
-            .orElseThrow(() -> EntityNotFoundException.forId("Applicant", applicantId));
+        Optional<Applicant> applicantOptional = applicantRepository.findById(userId);
+        Applicant applicant;
+        if (applicantOptional.isEmpty()) {
+            createApplicant(userId);
+            applicant = applicantRepository.findById(userId).orElseThrow();
```
MoSchmidt, line 78: This repository call is unnecessary and can be removed. I would rather return the Applicant directly from the createApplicant() method

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
```diff
@@ -466,4 +483,13 @@ public void deleteDocumentTypeOfDocuments(UUID applicationId, DocumentType docum
     public void renameDocument(UUID documentId, String newName) {
         documentDictionaryService.renameDocument(documentId, newName);
     }
+
+    void createApplicant(UUID userId) {
```
MoSchmidt, line 487: Pls specify the access modifier if there is no reason not to do it?

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
```diff
@@ -132,16 +149,16 @@ public ApplicationForApplicantDTO updateApplication(UpdateApplicationDTO updateA
         ApplicantDTO applicantDTO = updateApplicationDTO.applicant();
 
         Applicant applicant = applicantRepository.getReferenceById(applicantDTO.user().userId());
```
MoSchmidt, line 151: Not that important, but to make the code cleaner and improve the maintainability, I would suggest assigning the User to an extra variable and then just call user.set...()

MoSchmidt (changes requested):
Thanks for fixing this issue. Left 3 small code comments but functionality wise everything works as expected

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
Douglas-Sarah: Fixed 👍

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
Douglas-Sarah: Fixed 👍

marc-fett (approved):
Except for the comment from Moritz, for me the Code looks good. Thanks ! 👍🏼

az108 (approved):
LGTM

PR #536, opened 2025-08-15
`Development`: Add back button in Job Creation Form

ani-ib:
#### Client

### Motivation and Context

This change adds a “Back” button to the job creation form, allowing users to navigate to the previous page through the form panels more intuitively, and adds more clearly labeled buttons for navigating through the form. It improves UX by aligning the form navigation with user expectations for multi-step forms.

### Description

Added a translated “Back” button to the job creation form (job-creation-form.component.html) with appropriate styling and functionality.
• Refactored the layout header to accommodate the new button.
• Added translation keys for “step1” and “step2” in both English and German i18n files.
• Updated the form configuration to handle back-navigation between steps using the buttonGroupPrev with relevant labels and actions.

### Steps for Testing

Prerequisites:

1. Logged in as a professor

Steps:
1. Navigate to the Job Creation Form
2. Go through all the steps and verify that the bottom-left buttons for navigating backwards in the form are understandable for users and are translated correctly
3. Verify that the back button (that leads to the previous **page** is clear for users and is translated correctly

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### Screenshots


Kiara65 (changes requested):
Everything works as expected, thank you for integrating this. :) 
I just noticed that on the first page of the job creation form there are now two "Back" buttons — the new one in the upper right corner and the existing one in the lower left. Since they don’t look alike, this might be a bit confusing for users. I’d suggest either removing the lower-left button or aligning its style with the new one so it’s clear they have the same functionality. What is your opinion on that?
Additionally, it might be helpful to add a tooltip to the back button explaining that the form will be saved as a draft when pressed. Alternatively, this could also be addressed in a future PR by adding a confirmation dialog.

Issue #504, opened 2025-08-06
Professor landing page


marc-fett:
### Description

We need another landing page for professors. It should show the benefits and features of TUMApply. Additionally, we could add direct links to important information, tutorials on how to create a job, and much more.

Issue #485, opened 2025-08-04
Combine Application Creation and Login


marc-fett:
### Description

The login process should be integrated into the application creation process. The user can start to create an application but creates an account while doing so.

PR #547, opened 2025-08-17
`Development`: Add professor login

marc-fett:
#### General

#### Client

### Motivation and Context

This PR introduces the initial differentiation between applicant and professor login flows.
Until now, both shared the same entry points. With this change:
- Professors are redirected directly to the Keycloak SSO login.
- Applicants continue to use the standard login form.

⚠️ Disclaimer:
The Landing page itself is still identical for applicants and professors (only the header and navigation links differ).
A dedicated professor landing page will be implemented in Feature #504

### Description

- Updated header and navigation links depending on role (professor vs. applicant).
- Adjusted login handling:
 - Professor entry → redirect directly to Keycloak TUM login.
 - Applicant entry → show standard TUMApply login dialog.
- Ensured routing and guards work consistently.

### Steps for Testing

Steps:

1. Navigate to / (home page) as not-logged-in user.
2. Verify the header and links are applicant-specific (see screenshot below; default header, as known from before)
3. Click Login → the standard TUMApply login form appears.
4. Navigate to /professor (either via URL or via Button "Are you a professor?" as not-logged-in user.
5. Verify the header and links are professor-specific (see screenshot below; Professor text and different button text).
6. Click Login → you are redirected to the Keycloak login page.

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

### Screenshots

image

image

### DIFF
--- a/src/main/webapp/app/core/auth/keycloak.service.ts
+++ b/src/main/webapp/app/core/auth/keycloak.service.ts
@@ -20,13 +20,12 @@ export interface UserProfile {
 @Injectable({ providedIn: 'root' })
 export class KeycloakService {
   profile: UserProfile | undefined;
-  private refreshIntervalId: any;
-
   private readonly keycloak = new Keycloak({
     url: environment.keycloak.url,
     realm: environment.keycloak.realm,
     clientId: environment.keycloak.clientId,
   });
+  private refreshIntervalId: any;
 
   /**
    * Initializes the Keycloak client and determines login status.
@@ -107,7 +106,7 @@ export class KeycloakService {
         this.refreshIntervalId = undefined;
       }
       await this.keycloak.logout({
-        redirectUri: redirectUri ?? window.location.origin + '/',
+        redirectUri: redirectUri?.startsWith('http') ? redirectUri : window.location.origin + (redirectUri ?? '/'),
       });
     } catch (err) {
       console.error('Logout failed:', err);
--- a/src/main/webapp/app/app.routes.ts
+++ b/src/main/webapp/app/app.routes.ts
@@ -16,6 +16,13 @@ const routes: Routes = [
     title: 'landingPage.title',
   },
   {
+    path: 'professor',
+    canActivate: [UserRouteAccessService],
+    data: { authorities: [] },
+    loadComponent: () => import('./shared/pages/landing-page/landing-page.component').then(m => m.LandingPageComponent),
+    title: 'landingPage.title',
+  },
+  {
     path: '',
     canActivate: [UserRouteAccessService],
     data: { authorities: [] },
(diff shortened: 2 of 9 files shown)


marc-fett:
> Do we want to include a guard for logged-in users? Currently, if i'm logged in as a user and try to access /professor it will work and display Professor on the header (see image below) Screenshot 2025-08-19 at 18 02 35
> 
> Also according to the language guidelines we should use "Du" instead of "Sie"

Good question. I thought it would be good to have both landing pages publicly available. Maybe it feels a bit strange if the user can not access it when they are logged in as applicants, but can if they are not. But I removed the "Professor" text if the user is logged in as an applicant, since it seems a bit confusing. 

I replaced the "Sie with a "Du". Thank you for that 👍🏼 
Actually, we have many times "Sie"/"Ihr" in many texts. I opened a new issue for that: https://github.com/ls1intum/tum-apply/issues/566

Kiara65 (changes requested):
It works as expected on the test server, thank you for integrating this functionality.
I only noticed that the header is smaller compared to the one shown when the user is logged in, is this intended?
Also, will the professor landing page change in the future? At the moment, professors can click on Browse Positions and view the positions. However, when clicking Apply, the login dialog for applicants opens. If the professor landing page is not going to change, I would suggest showing an error confirmation dialog when a professor tries to apply. What do you think about that?

marc-fett:
> It works as expected on the test server, thank you for integrating this functionality. I only noticed that the header is smaller compared to the one shown when the user is logged in, is this intended? Also, will the professor landing page change in the future? At the moment, professors can click on Browse Positions and view the positions. However, when clicking Apply, the login dialog for applicants opens. If the professor landing page is not going to change, I would suggest showing an error confirmation dialog when a professor tries to apply. What do you think about that?

Thanks for testing. I changed the header. Now it has the same height when the user is logged in and logged out.

Regarding the professor landing page: This page will change to a professor-specific one. This will be done in [Issue 504](https://github.com/ls1intum/tum-apply/issues/504). Additionally, there won't be a login dialog anymore in the future when a user tries to apply for a job (Issue #549 and Issue #485)

Kiara65 (approved):

src/main/webapp/app/core/auth/keycloak.service.ts
```diff
@@ -20,13 +20,12 @@ export interface UserProfile {
...
 export class KeycloakService {
   profile: UserProfile | undefined;
-  private refreshIntervalId: any;
-
   private readonly keycloak = new Keycloak({
     url: environment.keycloak.url,
     realm: environment.keycloak.realm,
     clientId: environment.keycloak.clientId,
   });
+  private refreshIntervalId: any;
```
az108, line 28: please do not use any

az108 (changes requested):
small comment

PR #551, opened 2025-08-18
`Development`: Unified application form with local storage support

### LINKED ISSUE #549: Better UI Statemanagement for Applications
### Description

- server adjustment, such that the job overview has information for a logged in user about the state of the job
- fixes issue that the applicant can apply to the same job twice
- safer state management in the application creation form (through adjusted statemanagement in URL)
- allows expansion for login-less application

### Notes

_No response_

Douglas-Sarah:
closes #549 

#### General

#### Server

#### Client

## Motivation and Context

The current application form system uses two separate routes (`/application/create/{jobId}` and `/application/edit/{applicationId}`) which creates complexity in routing and state management. Additionally, unauthenticated users cannot use the application form as it always requires server-side persistence, limiting accessibility for users who want to draft applications before creating an account.

This change addresses several key issues:
- **Route complexity**: Multiple routes for essentially the same component with different modes
- **Accessibility barriers**: Unauthenticated users cannot access or use the application form
- **Data loss risk**: No draft persistence for users without accounts
- **User experience**: Seamless transition from anonymous drafting to authenticated submission

## Description

This PR introduces a unified application form system that consolidates routing and adds local storage support for unauthenticated users.

#### 1. **Unified Route Structure**
- **New route**: `/application/form?job={jobId}&application={applicationId}`
- **Backward compatibility**: Old routes redirect to the new format
- **Simplified logic**: Single component handles both create and edit modes

2. Enhanced Job Card Integration

- **Smart action buttons**: Job cards now reflect application state with appropriate actions
- **Prevent duplicate applications**: Authenticated users cannot apply twice to the same job
- **State-aware UI**: Different buttons for the cases "Apply", "Edit Draft", and "View Application"

#### 2. **Local Storage Integration**
- **Unauthenticated users**: Form data automatically saves to local storage every 3 seconds
- **Data persistence**: Drafts survive browser sessions and page refreshes
- **Seamless transition**: When users authenticate, local data can be migrated to server

#### 3. **Smart Application State Management**
- **Server-side validation**: Prevents editing of submitted applications (non-SAVED state)
- **Error handling**: Proper error messages and redirects for invalid states
- **Automatic routing**: Redirects to application detail page for non-editable applications

#### 4. **Enhanced User Experience**
- **Flexible entry points**: Support for both job-based creation and application-based editing
- **Progressive enhancement**: Works offline for unauthenticated users, full features for authenticated users
- **Data safety**: Clear separation between local and server storage

### Technical Implementation:

Technical Implementation:

- Modified `ApplicationCreationFormComponent` to handle unified routing
- Added local storage persistence with automatic save functionality
- Enhanced error handling for application state validation
- Updated route configuration with backward-compatible redirects
- Enhanced job card components with state-aware action buttons
- Added duplicate application prevention for authenticated users

### 1. **Test Job Cards**

- **Login as applicant**
- go to "Find Positions"
- verify that Jobs, where the applicant has already applied to, display a "View Application" button for the non-editable ApplicationStates (e.g. Withdrawn, Rejected, Sent), and a "Edit" button for the Saved - ApplicationState.
- verify that "View Application" routes to the application detail page
- verify that "Edit" routes to the application form, and that the application form has the data filled in from the existing application
- verify that "Apply" routes to the application form, and that the data is filled like it should be for a new application (applicant and user data - if already saved previously in database - displayed on page 1 and 2, at least empty page 3)

### 2. **Test Unauthenticated User Flow**

- **Logout from account**
- navigate to `/application/form?job={job_id}` (tip: you can copy a jobId from the URL of the job-detail-page for example)
- check that the email field is not disabled
- fill out some form fields
- wait for automatic save and then refresh page
- verify data persists after page refresh

### 3. **Test Authenticated User Flow**

- **Login as applicant**
- Navigate to `/application/form?job={job_id}`
- Fill out form fields
- Verify automatic server saves (check network tab)
- Refresh page and verify data persists

### 4. **Test Application State Validation**

- Create and submit an application (state = SENT)
- Try to navigate to `/application/form?application={submitted_app_id}`
- Verify error message displays
- Verify automatic redirect to `/application/detail/{application_id}`
- Test with other non-SAVED states (ACCEPTED, REJECTED, etc.)

### 5. **Test Error Scenarios**

- Navigate to `/application/form` (no parameters)
- Verify appropriate error message
- Navigate to `/application/form?job=invalid_id`
- Verify error handling for invalid job ID
- Navigate to `/application/form?application=invalid_id`
- Verify error handling for invalid application ID

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### Screenshots

The design of the job cards is being changed, so this design is temporary:
image
image

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,12 @@
   "schedule": [
     "every weekend"
   ],
+  "vulnerabilityAlerts": {
+    "enabled": true,
+    "labels": [
+      "security fix"
+    ]
+  },
   "packageRules": [
     {
       "groupName": "Weekly Dependency Updates",
@@ -22,26 +28,6 @@
       ],
       "schedule": [
         "every weekend"
-      ]
-    },
-    {
-      "groupName": "Security Updates",
-      "groupSlug": "security-updates",
-      "matchUpdateTypes": [
-        "major",
-        "minor",
-        "patch"
-      ],
-      "matchConfidenceLevels": [
-        "high"
-      ],
-      "labels": [
-        "security"
-      ],
-      "prPriority": 10,
-      "separateMultipleMajor": false,
-      "schedule": [
-        "at any time"
       ]
     }
   ]


src/main/webapp/app/job/job-overview/job-card/job-card.component.ts
```diff
@@ -29,6 +28,7 @@ export class JobCardComponent {
   workload = input<number | undefined>(undefined);
   startDate = input<string | undefined>('');
   relativeTime = input<string>('');
+  jobCard = input<JobCardDTO | undefined>(undefined);
```
marc-fett, line 31: since you now give the whole job to the component, can't we remove the other props?

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
```diff
@@ -58,9 +58,21 @@ public class ApplicationService {
...
+        if(userId == null) {
+            Application application = new Application();
+            application.setJob(job);
+            application.setState(ApplicationState.SAVED);
+            return ApplicationForApplicantDTO.getFromEntity(application);
+        }
+
+        Application a = applicationRepository.getByApplicant_User_UserIdAndJob_JobId(userId, jobId);
+        if (a != null) {
+            return ApplicationForApplicantDTO.getFromEntity(a);
```
marc-fett, line 75: shouldn't we create a new application here? otherwise we would return null

src/main/java/de/tum/cit/aet/job/dto/JobCardDTO.java
```diff
@@ -25,6 +29,34 @@ public JobCardDTO(
...
+            professorName,
+            applicationId,
+            applicationState,
+            workload,
+            startDate,
+            UiTextFormatter.getRelativeTimeLabel(createdAt)
+        );
+    }
+
+        public JobCardDTO(
```
marc-fett, line 52: why do we need two jobCardDTOs

src/main/java/de/tum/cit/aet/job/repository/JobRepository.java
```diff
@@ -19,103 +19,123 @@
...
-        """
-    )
+    @Query("""    
+            SELECT new de.tum.cit.aet.job.dto.JobCardDTO(
+                j.jobId,
+                j.title,
+                j.fieldOfStudies,
+                j.location,
+                CONCAT(j.supervisingProfessor.firstName, ' ', j.supervisingProfessor.lastName),
+                (SELECT a1.applicationId
```
marc-fett, line 80: You are querying the latest Application twice (for applicationId and state). I suggest introducing a single LEFT JOIN to fetch the latest application once and reuse it for both fields. This will improve readability and query performance.

```LEFT JOIN Application a
  ON a.job = j
 AND a.applicant.userId = :userId
 AND a.createdAt = (
   SELECT MAX(a2.createdAt)
   FROM Application a2
   WHERE a2.job = j
     AND a2.applicant.userId = :userId
 )```

Then use a.applicationId and a.state directly.

File: src/main/java/de/tum/cit/aet/job/repository/JobRepository.java [hunk 5]
```diff
@@ -19,103 +19,123 @@
...
+                                    FROM Application a2 
+                                    WHERE a2.job = j 
+                                    AND a2.applicant.userId = :userId)),
+                j.workload,
+                j.startDate,
+                j.createdAt
+            )
+            FROM Job j
+            WHERE j.state = :state
+            AND (:title IS NULL OR j.title LIKE %:title%)
```
marc-fett, line 102: The LIKE %:param% patterns can be improved for both safety and consistency. Consider using LOWER() with CONCAT to ensure case-insensitive search.

```AND (:title IS NULL OR LOWER(j.title) LIKE LOWER(CONCAT('%', :title, '%')))```

Same could be applied for the other filters

File: src/main/java/de/tum/cit/aet/job/repository/JobRepository.java [hunk 6]
```diff
@@ -19,103 +19,123 @@
...
+            AND (:fieldOfStudies IS NULL OR j.fieldOfStudies LIKE %:fieldOfStudies%)
+            AND (:location IS NULL OR j.location = :location)
+            AND (:professorName IS NULL OR
+                CONCAT(j.supervisingProfessor.firstName, ' ', j.supervisingProfessor.lastName) LIKE %:professorName%)
+            AND (:workload IS NULL OR j.workload = :workload)
+            ORDER BY
+                CASE WHEN :sortDirection = 'ASC'  AND :sortBy = 'professorName' THEN CONCAT(j.supervisingProfessor.firstName, ' ', j.supervisingProfessor.lastName) END ASC,
```
marc-fett, line 109: To ensure deterministic results, please add a secondary sort (e.g. j.createdAt DESC) when sorting by professor name. Otherwise, results may be inconsistent if multiple professors have the same name.

src/main/java/de/tum/cit/aet/job/service/JobService.java
```diff
@@ -201,6 +202,12 @@ public JobDetailDTO getJobDetails(UUID jobId) {
      * @return a page of {@link JobCardDTO} matching the criteria
      */
     public Page<JobCardDTO> getAvailableJobs(PageDTO pageDTO, AvailableJobsFilterDTO availableJobsFilterDTO, SortDTO sortDTO) {
+        UUID userId = null;
+        try {
```
marc-fett, line 206: please add a new method in currentUserService instead of catching all "AccessDeniedExceptions" by default

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -383,4 +502,54 @@ export default class ApplicationCreationFormComponent {
   onPage3ValidityChanged(isValid: boolean): void {
     this.page3Valid.set(isValid);
   }
+
+  private saveToLocalStorage(): void {
```
marc-fett, line 506: please add a separate service for all local storage related functionality since we will probably need this in multiple components

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -260,57 +259,160 @@ export default class ApplicationCreationFormComponent {
...
+    return application;
   }
 
   async performAutomaticSave(): Promise<void> {
     if (this.savingState() === SavingStates.SAVING) {
-      await this.sendCreateApplicationData(this.applicationState(), false);
+      if (this.useLocalStorage()) {
+        this.saveToLocalStorage();
+      } else {
+        await this.sendCreateApplicationData(this.applicationState(), false);
```
marc-fett, line 393: please surround with a try catch and only set the saved state when no it was successful

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -27,27 +27,15 @@ import ApplicationCreationPage2Component, {
...
-type ApplicationState = (typeof ApplicationStates)[keyof typeof ApplicationStates];
-
 const SavingStates = {
   SAVED: 'SAVED',
   SAVING: 'SAVING',
 } as const;
 
 type SavingState = (typeof SavingStates)[keyof typeof SavingStates];
 
+const LOCAL_STORAGE_KEY = 'application_draft';
```
marc-fett, line 37: If we want to have the possibility to apply for multiple jobs, We need to use a storage key based on the application and/or job id. If you have multiple drafts, they get overwritten. In this case the load from local storage needs to be adapted.

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -95,42 +84,51 @@ export default class ApplicationCreationFormComponent {
...
   savingBadgeCalculatedClass = computed<string>(
     () =>
       `flex flex-wrap justify-around content-center gap-1 ${this.savingState() === SavingStates.SAVED ? 'saved_color' : 'unsaved_color'}`,
   );
-  mode: ApplicationFormMode = 'create';
+
   page1Valid = signal<boolean>(false);
   page2Valid = signal<boolean>(false);
   page3Valid = signal<boolean>(false);
   savingTick = signal<number>(0);
```
marc-fett, line 110: savingTick is not used

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -383,4 +502,54 @@ export default class ApplicationCreationFormComponent {
...
   }
+
+  private saveToLocalStorage(): void {
+    const applicationData = {
+      page1: this.page1(),
+      page2: this.page2(),
+      page3: this.page3(),
+      applicationId: this.applicationId(),
+      jobId: this.jobId(),
+      timestamp: new Date().toISOString(),
```
marc-fett, line 513: you save the timestamp but do not use it. We can say e.g. that a 30 days old draft is not valid anymore

```const TTL_MS = 30 * 24 * 60 * 60 * 1000;
if (Date.now() - (data.timestamp ?? 0) > TTL_MS) {
  localStorage.removeItem(this.storageKey());
  return false;
}```

And we can always use the latest draft (if there are multiple ones). In the future we could also show to the user when the draft was edited the last time.

File: src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts [hunk 13]
```diff
@@ -260,57 +259,160 @@ export default class ApplicationCreationFormComponent {
...
+          summary: 'Error',
+          detail: 'This application cannot be edited as it has already been submitted or is in a non-draft state.',
+        });
+        await this.router.navigate(['/application/detail', application.applicationId]);
+        return;
+      }
+      this.applicationId.set(application.applicationId ?? '');
+
+      // Update URL to include the new applicationId
+      await this.router.navigate([], {
```
marc-fett, line 372: It would be nice to update the page based on the step the user is currently on but this could also happen in a future PR.

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.html
```diff
@@ -11,7 +11,13 @@
 />
 <ng-template #panel1>
   <div class="panel">
-    <jhi-application-creation-page1 class="w-full" [(data)]="page1" (valid)="onPage1ValidityChanged($event)" (changed)="onValueChanged()" />
+    <jhi-application-creation-page1
+      class="w-full"
+      [(data)]="page1"
+      (valid)="onPage1ValidityChanged($event)"
+      (changed)="onValueChanged()"
+      [disableEmail]="!useLocalStorage()"
```
marc-fett, line 19: please disable based on if the user is already logged in. Then we do not need to handover disableEmail everywhere

marc-fett (commented):
Sorry accidentially sent the feedback without a comment. 

Thank you very much for the implementation! 👍🏼 
I added some comments to improve the performance a bit and some ideas for the future. 

While testing, I found two bugs:
1. When no application and job is provided, no error is thrown.
1. When creating a new application or editing an old one, Find Positions and My Applications is selected in the sidebar.

src/main/webapp/app/job/job-overview/job-card/job-card.component.ts
Douglas-Sarah: I have replaced jobCard with applicationId and applicationState 👍

src/main/java/de/tum/cit/aet/application/service/ApplicationService.java
Douglas-Sarah: For the case that an application already exists, we return that application. Otherwise we create a new one. I think that makes more sense, because it makes the function more resilient, and we only have to execute one server call when opening the application form page. But we can rename the function for example

src/main/java/de/tum/cit/aet/job/dto/JobCardDTO.java
Douglas-Sarah: For the database functionality, it was useful having a constructor which I could give the ApplicationState as a String. So we have 2 constructors, one which can get the ApplicationState as an Enum, and one for the ApplicationState as a string.

To add more context - the JobState does not have the same problem in the JobRepository, because it has a simple constructor. The constructor of the ApplicationState includes information about who the authorization to access an application at each state. This is an artefact from the the server planning stage. 
You don't plan on using it for authorization, right? Because we could remove it theoretically. Maybe in a separate PR to not blow up the line count further?
I saw as well that in the EmailType, the UserRole is also used as an attribute.

marc-fett: I do not need it in the authorization. If it still makes sense, we can keep it, but it seems a bit inefficient to have almost the same constructors except for one variable. Maybe we can change the JobRepository to return an enum?

Douglas-Sarah: I used mapping within the query, which fixed this issue and the one below 👍 So we only have one constructor again

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: Changed by now 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: Removed 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: Added a TTL logic for a 30 day validity duration 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: I agree 👍 But i see this as a future PR

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.html
Douglas-Sarah: Now disabled through userservice👍

src/main/java/de/tum/cit/aet/job/repository/JobRepository.java
Douglas-Sarah: Thanks, this was already implemented in your suggestion for the query

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: There is already an Issue for that matter- there will be a third saving state implemented which will show when the saving try failed (https://github.com/ls1intum/tum-apply/issues/594)
So this will come in a separate PR

src/main/java/de/tum/cit/aet/job/repository/JobRepository.java
Douglas-Sarah: Added the secondary sort 👍

marc-fett (changes requested):
Thanks for the fix. I found a new problem and added some comments. The problem now is that when there is no application yet, you get automatically redirected to the landing page. It should be possible to add a new application and the browser should not show an error in this case.

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -245,75 +239,179 @@ export default class ApplicationCreationFormComponent {
...
   private applicationResourceService = inject(ApplicationResourceService);
   private accountService = inject(AccountService);
   private route = inject(ActivatedRoute);
+  private router = inject(Router);
   private toastService = inject(ToastService);
+  private localStorageService = inject(LocalStorageService);
 
-  constructor() {
-    this.init();
+  private initEffect = effect(() => {
```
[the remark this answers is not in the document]
Douglas-Sarah: I understand that calling the effect twice is not ideal, which is why i propose this:
```
private initEffect = effect(() => {
    if (!untracked(() => this.initCalled())) {
      this.initCalled.set(true);
      this.init();
    }
  });
```
I question the use of the platformId in your example- we do not use server side rendering. Is there any reason to check whether we are in a browser?

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -534,25 +527,28 @@ export default class ApplicationCreationFormComponent {
...
+    this.localStorageService.clearApplicationDraft(this.applicationId(), this.jobId());
+  }
+
+  private showInitErrorMessage(summary: string, detail: string): void {
+    queueMicrotask(() => {
+      this.toastService.showError({
+        summary,
+        detail,
+      });
+      setTimeout(() => void this.router.navigate(['']), 3000);
```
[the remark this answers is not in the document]
Douglas-Sarah: Changed 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -245,75 +239,179 @@ export default class ApplicationCreationFormComponent {
...
 
   async performAutomaticSave(): Promise<void> {
     if (this.savingState() === SavingStates.SAVING) {
-      await this.sendCreateApplicationData(this.applicationState(), false);
+      if (this.useLocalStorage()) {
+        this.saveToLocalStorage();
+      } else {
+        await this.sendCreateApplicationData(this.applicationState(), false);
+      }
       this.savingState.set(SavingStates.SAVED);
```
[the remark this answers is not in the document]
Douglas-Sarah: I have mentioned this here https://github.com/ls1intum/tum-apply/pull/551#discussion_r2301993008 -  having a third saving state for failed state is already planned and the automatic save will get a do over, a shared service most likely.
I have changed the code to not update the state to SAVED in case of an error nonetheless.

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -245,75 +239,179 @@ export default class ApplicationCreationFormComponent {
...
+  });
 
-    effect(() => {
-      const intervalId = setInterval(() => {
-        this.performAutomaticSave();
-      }, 3000);
-      return () => clearInterval(intervalId);
-    });
-  }
+  private automaticSaveEffect = effect(() => {
```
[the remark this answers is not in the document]
Douglas-Sarah: I agree that your solution is more in line with the reactive use of signals and reduces overhead.

I would suggest doing this update within the automatic save do-over PR. This change would need extensive testing and fits more into the other PR i mentioned. 
(For context: At the moment the automatic save function checks every 3 seconds for changes, but does not perform a save if there is no change)

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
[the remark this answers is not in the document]
marc-fett: This is only for testing reasons since you do not have the possibility to access the local storage there. But this was just a suggestion.

marc-fett (changes requested):
Thanks for the fixes. The code LGTM. I found one more issue while testing again:

I can enter http://localhost:4200/application/form?job=invalidId and get no error.

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -245,75 +239,179 @@ export default class ApplicationCreationFormComponent {
...
+        relativeTo: this.route,
+        queryParams: { job: jobId, application: application.applicationId },
+        queryParamsHandling: 'merge',
+      });
+    } catch (error) {
+      this.toastService.showError({
+        summary: 'Error',
+        detail: 'Failed to create application. Please try again.',
+      });
+      console.error('Failed to create application:', error);
```
az108, line 378: please get rid of the console.error

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -348,27 +446,46 @@ export default class ApplicationCreationFormComponent {
...
         location.back();
       }
     } catch (err) {
       const httpError = err as HttpErrorResponse;
-      this.toastService.showError({ summary: 'Error', detail: 'Failed to save application: ' + httpError.statusText });
+      this.toastService.showError({
+        summary: 'Error',
+        detail: 'Failed to save application: ' + httpError.statusText,
+      });
       console.error('Failed to save application:', err);
```
az108, line 467: same here remove the console.log

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
```diff
@@ -386,4 +503,63 @@ export default class ApplicationCreationFormComponent {
...
+    this.localStorageService.clearApplicationDraft(this.applicationId(), this.jobId());
+  }
+
+  private showInitErrorMessage(summary: string, detail: string): void {
+    queueMicrotask(() => {
+      this.toastService.showError({
+        summary,
+        detail,
+      });
+      setTimeout(() => void this.router.navigate(['/job-overview']), 3000);
```
az108, line 562: why do we need to display the toast like this?

az108 (changes requested):
small comments

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: removed 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
Douglas-Sarah: removed 👍

src/main/webapp/app/application/application-creation/application-creation-form/application-creation-form.component.ts
rabeatwork: @az108, we added a comment for now to no longer delay this PR. However, we should investigate this issue another time.

rabeatwork (approved):

PR #564, opened 2025-08-19
`General`: Fix date selection scroll behavior

Bofan-Zhu:
#### General

### Motivation and Context

- Date picker unexpectedly closes when scrolling inside the date picker component

### Description

- Change to how date picker is rendered -> date picker stays open if user accidentally scrolls inside the date picker component, closes if the user scrolls outside of the date picker

### Steps for Testing

Prerequisites:

1. Log in to TumApply as Professor/Applicant
2. View Job Position or apply to Job and check forms with date picker

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

### DIFF
--- a/src/main/webapp/app/shared/components/atoms/datepicker/datepicker.component.html
+++ b/src/main/webapp/app/shared/components/atoms/datepicker/datepicker.component.html
@@ -27,6 +27,7 @@
     (onHide)="isCalendarOpen = false"
     (onShow)="isCalendarOpen = true"
     [(ngModel)]="modelDate"
+    [appendTo]="'body'"
     [placeholder]="shouldTranslate() && placeholder() ? (placeholder() ?? '' | translate) : placeholder()"
     [disabled]="disabled()"
     [showButtonBar]="true"


marc-fett (approved):
Thanks for implementing. 👍🏼

az108 (approved):

src/main/webapp/app/shared/components/atoms/select/select.component.scss
```diff
@@ -80,3 +80,10 @@
     border: 0.1rem solid var(--p-primary-color);
   }
 }
+
+:root {
```
az108, line 84: the :root is unneccesary here as you use !important . Ideally we can get rid of !important aswell as it should only be used in cases where its not avoidable. Try it without and if it doesnt work because of some prime ng stylings then we can even consider moving this to the global.sccss as select.component is used on several pages and its an atomic component

Bofan-Zhu: Moved it to the global file to avoid potential redundancy

az108 (approved):

PR #642, opened 2025-09-03
`Development`: Improve job form layout

ani-ib:
#### Client

### Motivation and Context

The job creation form layout had usability issues, with the “Start Date” and “Application Deadline” fields placed in different rows. This made the form feel disjointed. Feedback from last sprint review also mentioned replacing the 5 year maximum limit for contract duration with a 6 year limit. The job creation form also had an unwanted shadow surrounding it, which has now been replaced by a border.

### Description

- Moved the “Start Date” field next to the “Application Deadline” field within the same row for a more logical form grouping.
- Changed the upper limit for “Contract Duration” from 5 to 6 years.
- Updated the tooltip text for the “Contract Duration” field in both English and German translations to reflect the correct 6-year limit.
- Removed the **shadow-box** property around the job creation form  and replaced it with a border.

### Steps for Testing
Steps:

1. Log in to TumApply as a professor
e.g. 
username: professor1@tumapply.local
password: professor
3. Open the Job Creation Form
4. Verify that the **application deadline** and **start date** fields are in the same row
5. Verify that the contract duration is constricted to maximum 6 years now and not 5.
6. Verify that the tooltip has been updated to inform users that they can only enter max 6 years into the contract duration field.
7. Verify that (when making the screen size bigger) there is a simple border that appears around the form, and no longer a shadow.

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### Screenshots

Updated Job Creation Form with Application Deadline and Start Date in same row **and** no shadow around the form's border.
image
Updated Contract Duration tooltip with 6 instead of 5:
image


Abi107717 (approved):
LGTM, thank you for implementing

Just a small suggestion: Maybe we could show an error message when a number greater than 6 is entered, so it’s more intuitive why the number gets automatically changed to 6.
But this could be a separate PR. :)

Kiara65 (approved):
LGTM, thank you 😊

ani-ib:
> LGTM, thank you for implementing
> 
> Just a small suggestion: Maybe we could show an error message when a number greater than 6 is entered, so it’s more intuitive why the number gets automatically changed to 6. But this could be a separate PR. :)

Thanks for the feedback, this is a good point. Currently by default, the atomic number component always has a minimum and maximum value, so we would have to adjust this component as well in order to implement this. I will add this as an issue to the sprint board 😊

ani-ib:
Thanks for your feedback @rabeatwork I've implemented your change request

rabeatwork (approved):

PR #723, opened 2025-09-15
`General`: Remove production website ribbon

az108:
### Motivation and Context

As the website will have its kickoff today, we want to get rid of the demo ribbon when real test users will use the website.

### Description

Removed Page Ribbon for the Production Environment.

### Steps for Testing

Can only be verified on Production or by simulating Production environment locally.

#### Code Review

- [ ] Code Review 1


Abi107717 (approved):
LGTM

PR #820, opened 2025-09-25
bugfix: review dialog default sort

MoSchmidt:
### Motivation and Context

When directly going to the review page, a server error was thrown. This was caused by the client sending the wrong sort field (createdAt), which was set as a default. This PR adapts the sort field to "appliedAt" so the validation passes again.

### Steps for Testing

- [ ] directly go to the review page and see that all applications are loaded correctly

Prerequisites:

1. Log in to TumApply as a Professor


az108 (approved):

Kiara65 (approved):
It is working again now, thank you! 😄

PR #1052, opened 2025-10-19
`Development`: Add internal comment server tests

MoSchmidt:
### Description

- added integration tests for the `InternalCommentResource` and `InternalCommentService`

### Steps for Testing

- check code quality and verify tests are passing

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

### Test Coverage

grafik

grafik


Bofan-Zhu (changes requested):
Thank you for applying the changes 😊. Can you also resolve the codacy errors? Some Tests in the InternalCommentResourceTest do not include assert()

MoSchmidt:
> Thank you for applying the changes 😊. Can you also resolve the codacy errors? Some Tests in the InternalCommentResourceTest do not include assert()

The problem is that the MvcTesClient does contain the assertions. However, codacy does not recognize them and thus throws the error. But I added some null-checks anyways so it is clearer that the expected response is null (empty body)

PR #1528, opened 2025-12-14
`General`: Use tooltip as fallback for button arialabel

Bofan-Zhu:
#### General

### Motivation and Context

Buttons currently use icon name as fallback for arialabels. This should be changed as icon names are not very descriptive (e.g. eye, xmark). Since all only icon buttons should have a tooltip, we can use the tooltip as the arialabel instead.

### Description

Change fallback to tooltip instead of icon

### Steps for Testing

Prerequisites:

1. Log in to TumApply as Professor/Applicant
2. Check buttons have ariaLabel by either checking accessibility tree or enabling voice over

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### DIFF
--- a/src/main/webapp/app/shared/components/atoms/button/button.component.html
+++ b/src/main/webapp/app/shared/components/atoms/button/button.component.html
@@ -9,7 +9,7 @@
 </div>
 <ng-template #button>
   <p-button
-    [ariaLabel]="label() === undefined ? icon() : undefined"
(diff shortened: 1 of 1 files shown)


Kiara65 (approved):
Tested it on the testserver, accessibility tree is showing the aria-label for buttons with undefined labels, thank you for adding this 🥳

PR #1754, opened 2026-01-20
`General`: Adjust button alignment

Bofan-Zhu:
#### General

### Motivation and Context

Some buttons are misaligned and do not seem like they belong to their content.

### Description

This PR aligns all buttons on the professor page to their content (top right of the content)

### Steps for Testing

Prerequisites:

1. Log in to TumApply as Professor
2. Check alignment of buttons

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### Screenshots

image
image
image

### DIFF
--- a/src/test/webapp/app/shared/components/molecules/search-filter-sort-bar.spec.ts
+++ b/src/test/webapp/app/shared/components/molecules/search-filter-sort-bar.spec.ts
@@ -282,7 +282,7 @@ describe('SearchFilterSortBar', () => {
         totalRecords: 10,
       });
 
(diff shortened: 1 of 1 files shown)


Bofan-Zhu:
> LGTM! I tested it on the test server and also had a look on the Code. One thing I noticed is that the second filter selector is longer than the first one and maybe if we make the second one in the same length as the first one we get even a better alignment image

Thanks for reviewing. I would change that in another PR. I think one is wider than the other because the sort displays the current sorted state but the multi select does not display the currently selected options and they both have different width (different components)

az108 (approved):

PR #2242, opened 2026-04-03
`Development`: Improve input validation and access control

krusche:
## Summary

- **Sanitize rich-text input fields** before persisting to the database. All HTML content from user-editable fields (application motivation, special skills, projects, job descriptions) is now cleaned using the existing `HtmlSanitizer` utility to ensure only safe formatting tags are stored. This covers all write paths including translations.
- **Harden template engine configuration** by restricting the FreeMarker `Configuration` at startup to prevent arbitrary class instantiation from user-supplied templates. Normal template variable interpolation continues to work as expected.
- **Add authorization check for research group member management** to verify the requesting user belongs to the target research group before allowing member additions. Prevents users from modifying groups they are not part of.

## Test plan

- [x] New unit test verifies HTML sanitization strips unsafe tags (`<script>`, `<iframe>`, ``) while preserving safe formatting (`<p>`, `<b>`, `<i>`, `<a>`)
- [x] New unit test verifies template engine rejects class instantiation directives while allowing normal variable substitution
- [x] New unit test verifies member addition is rejected when the user does not belong to the target group
- [x] New unit test verifies member addition succeeds when the user belongs to the target group
- [x] All existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/de/tum/cit/aet/core/util/HtmlSanitizer.java
+++ b/src/main/java/de/tum/cit/aet/core/util/HtmlSanitizer.java
@@ -22,6 +22,9 @@ public class HtmlSanitizer {
     /** Allowed CSS class pattern: only Quill formatting classes (alignment, indentation). */
     private static final Pattern ALLOWED_CLASS_PATTERN = Pattern.compile("ql-align-(center|right|justify)|ql-indent-[1-8]");
 
+    /** Allowed values for the data-list attribute on &lt;li&gt; elements (Quill 2.x list types). */
+    private static final Set<String> ALLOWED_DATA_LIST_VALUES = Set.of("bullet", "ordered");
+
     /**
      * Sanitizes generic HTML input using a base safelist. Allows basic formatting and anchor tags.
      * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/de/tum/cit/aet/job/service/JobService.java
```diff
@@ -202,16 +206,16 @@ public JobDetailDTO getJobDetails(UUID jobId) {
         return new JobDetailDTO(
             job.getJobId(),
             job.getSupervisingProfessor().getFirstName() + " " + job.getSupervisingProfessor().getLastName(),
-            job.getSupervisingProfessor().getResearchGroup(),
+            ResearchGroupSummaryDTO.getFromEntity(job.getSupervisingProfessor().getResearchGroup()),
```
CelineLahnor: Job has its own researchGroup foreign key, but the new JobDetailDTO is populated from job.getSupervisingProfessor().getResearchGroup() instead. If a professor's membership changes after a job is created, the detail endpoint will expose the wrong group or null even though the job still belongs to the original group, which also breaks the new summary DTO and belongsToResearchGroup logic on the frontend.

My suggestion would be to build JobDetailDTO.researchGroup from job.getResearchGroup() instead of job.getSupervisingProfessor().getResearchGroup() so the response reflects the job’s persisted ownership, not the professor’s current membership.

Cathy0123456789 (approved):
Thank you, works now on TS, also Job Export renders correctly and adding research group members still works. Testing locally, emails are still sent

PR #2306, opened 2026-04-15
`Development`: Add compliance text highlighting  client side

### LINKED ISSUE #2064: Add compliance client-side
### Description

Each issue is highlighted in its respective category color (red / yellow). Clicking a highlighted passage opens an overlay panel with the violated article, explanation and recommended action.

### Tasks

- [x] Highlighting
- [x] Compliance in Score calculation 

### Acceptance Criteria

_No response_

ge94zec:
#### General

#### Client

### Motivation and Context

This PR introduces client-side compliance text highlighting for job descriptions. The goal is to make detected compliance issues immediately visible and understandable to users while editing job postings.

Closes: https://github.com/ls1intum/tum-apply/issues/2064

### Description

Server-side: Compliance issues are extended with a language attribute to correctly associate issues with job descriptions across language switches. The database schema (Liquibase), ComplianceIssue DTO, and AiService were updated accordingly. The AI service now explicitly assigns the detected language to each compliance issue after analysis.

Client-side: The job creation form introduces a complianceIssues signal to manage compliance issues for both languages independently. Issues are filtered by the active editor language and dynamically applied as highlights. A  highlightEffect ensures highlights are applied whenever the language switches or new issues arrive.
A custom Quill blot (HighlightBlot) renders highlighted text spans using Tailwind utility classes, visually distinguishing critical AGG violations and transparency issues via different colors. 

TranslateText.st was updated to return only the translated text, ensuring compliance and gender bias issues introduced during manual editing are preserved and not silently corrected by the translation.

Fixed a bug where a score of 0 was not displayed due to an early return in the animation effect when animatedScore was undefined but targetScore was 0.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply
2. Navigate to Create new Position
3. Check if non compliant text phrases get highlighted 

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [x] Test 1

### Screenshots

Bildschirmfoto 2026-04-18 um 00 20 16
Bildschirmfoto 2026-04-18 um 00 16 29
Bildschirmfoto 2026-04-18 um 00 16 23

#### Client

| job-creation-form.component.ts | 62.95% | 1092 | 124 | 11.4 |
| ai-score-ring.component.ts | 88.67% | 73 | 3 | 4.1 |
| editor.component.ts | 80.79% | 306 | 56 | 18.3 |

#### Server

| ComplianceIssue.java | 0.00% | 26 |
| AiService.java | 8.70% | 224 |
| ComplianceScoreService.java | 85.71% | 81 |
| AiResource.java | 50.00% | 70 |
| Job.java | 100.00% | 81 |
| JobDTO.java | 100.00% | 33 |
| JobFormDTO.java | 95.45% | 61 |
| JobService.java | 78.33% | 310 |

_Last updated: 2026-04-19 14:53:25 UTC_


src/main/java/de/tum/cit/aet/ai/service/AiService.java
```diff
@@ -35,6 +32,8 @@
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Service;
 import reactor.core.publisher.Flux;
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.ObjectMapper;
```
Cathy0123456789: These imports are unused

src/main/java/de/tum/cit/aet/ai/service/AiService.java
```diff
@@ -17,10 +17,7 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
Cathy0123456789: Is there a reason to combine them?

src/main/resources/config/liquibase/changelog/00000000000033_add_ai_compliance_to_jobs.xml
```diff
@@ -21,6 +21,7 @@
       <column name="article" type="varchar(255)"/>
       <column name="explanation" type="text"/>
       <column name="action" type="varchar(50)"/>
+      <column name="language" type="varchar(5)"/>
```
Cathy0123456789: We save language normally with VARCHAR(10), keeping it consistent might be good

src/main/resources/prompts/TranslateText.st
```diff
@@ -8,7 +8,7 @@ INPUT TEXT (HTML):
 OUTPUT CONSTRAINTS - CRITICAL:
 1. Fidelity: You must preserve the exact formatting of the original text, including line breaks, bullet points, and paragraphs.
 2. Style: Provide a natural, grammatically correct translation that mirrors the professional tone of the source. Prioritize functional equivalence over literal word-for-word translation.
-3. Integrity: Use gender-inclusive language appropriate for Higher Education (e.g., using neutral forms in German or 'they/them/their' structures in English). Avoid words with forbidden stems: [{nonInclusiveWords}]. Rather use alternatives from: [{inclusiveWords}].
+3. Integrity: Use language appropriate for Higher Education (e.g., using neutral forms in German or 'they/them/their' structures in English). Translate ALL content literally and completely. Your task is translation only.
```
Cathy0123456789: Translating content literally led to problems normally, we rather want grammatically correct content depending on the context

Cathy0123456789 (changes requested):
Thank you, small comments

src/main/java/de/tum/cit/aet/ai/service/AiService.java
ge94zec: This came from merging an older version of the branch. I’ll clean it up :)

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
```diff
@@ -11,13 +11,64 @@ import { GenderBiasAnalysisResponse } from 'app/generated/model/gender-bias-anal
...
+  // HTML tag that will wrap the highlighted text
+  static tagName = 'span';
+  // CSS class that allows Quill to identify elements in DOM
+  static className = 'compliance-highlight';
+
+  // Tailwind classes applied to every highlighted text span
+  static utilityClasses = [
+    'border-b-2',
+    '[border-bottom-style:solid]',
+    '[border-bottom-color:var(--highlight-color)]',
```
CelineLahnor: Here as well maybe it would make sense to add semantic tokens for the highlight colors in general

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1520,8 +1556,19 @@ export class JobCreationFormComponent {
...
       // 2) Send the description to the analysis endpoint (persists score on the backend)
-      await firstValueFrom(this.aiApi.analyzeJobDescriptionForCompliance(lang, jobForm));
+      const compliance = await firstValueFrom(this.aiApi.analyzeJobDescriptionForCompliance(lang, jobForm));
       this.lastAnalyzedText = descriptionText;
+      // Keep issues from other languages, but replace all issues for the current language with the latest analysis.
+      const otherLang = lang === 'en' ? 'de' : 'en';
+      const existingLang = this.complianceIssues().filter(issue => issue.language === otherLang);
+
+      const incomingLang = compliance.map(issue => {
+        const copy: any = structuredClone(issue);
```
CelineLahnor: Could we avoid the any here?

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
```diff
@@ -11,13 +11,64 @@ import { GenderBiasAnalysisResponse } from 'app/generated/model/gender-bias-anal
...
+    'rounded-[var(--border-radius-xs)]',
+    '[box-decoration-break:clone]',
+    '[-webkit-box-decoration-break:clone]',
+    'transition-colors',
+    'duration-150',
+    'cursor-pointer',
+    'hover:[background-color:var(--highlight-bg)]',
```
CelineLahnor: Do we have a semantic token for this? Would it make sense to add one?

CelineLahnor (changes requested):
Thank you code looks already good to me I just left some small questions in the code :)

CelineLahnor:
Can you please create a follow-up issue to add the missing test coverage introduced by this PR, so it does not get forgotten?

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -30,7 +33,7 @@ OUTPUT:
 - No prose.
 - One object per issue.
 - Object fields must be exactly:
-  id, text, category, article, explanation, action
+  id, text, category, article, explanation, action, language
```
Cathy0123456789: Does it make sense for the AI to set the language if we anyway overwrite it in AiService?

src/main/resources/prompts/TranslateText.st
```diff
@@ -8,7 +8,7 @@ INPUT TEXT (HTML):
 OUTPUT CONSTRAINTS - CRITICAL:
 1. Fidelity: You must preserve the exact formatting of the original text, including line breaks, bullet points, and paragraphs.
 2. Style: Provide a natural, grammatically correct translation that mirrors the professional tone of the source. Prioritize functional equivalence over literal word-for-word translation.
-3. Integrity: Use gender-inclusive language appropriate for Higher Education (e.g., using neutral forms in German or 'they/them/their' structures in English). Avoid words with forbidden stems: [{nonInclusiveWords}]. Rather use alternatives from: [{inclusiveWords}].
+3. Integrity: Use language appropriate for Higher Education (e.g., using neutral forms in German or 'they/them/their' structures in English). Prioritize grammatical correctness and readability in the target language over literal word mapping. Your task is translation only.
```
Cathy0123456789: Thinking about it a second time, why are we removing the "gender-inclusive" part? Even if the original text is not gender-inclusive, we do want to achieve this in the translated text nonetheless, I think? @az108 

I remember adding this actually after having gotten this feedback. If we take a professor for example who doesn't care to correct the compliance issues, we have at least one language that is not gender-exclusive this way

Cathy0123456789 (changes requested):
Thank you, small questions

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: unfortunately I can not because of the create method call in editor.component.ts

src/main/resources/prompts/TranslateText.st
az108: yes we should keep the gender-inclusive instruction

src/main/java/de/tum/cit/aet/ai/domain/ComplianceIssue.java
az108, line 1: can you create a domain folder and move this file to there please, its not fully a dto

src/main/resources/prompts/AnalyzeComplianceText.st
az108: i agree please remove the language from here if you set it manually afterwards, and its less error prone and saves ai tokens so that it focuses on more relevant things

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
az108: it would make sense to add something like this in the token_scss and use that

```
  --color-compliance-critical-border: var(--color-negative-DEFAULT);                                                                                                                                                                   
  --color-compliance-critical-bg: color-mix(in srgb, var(--color-negative-DEFAULT) 18%, transparent);                                                                                                                                  
  --color-compliance-warning-border: var(--color-warning-DEFAULT);                                                                                                                                                                     
  --color-compliance-warning-bg: color-mix(in srgb, var(--color-warning-DEFAULT) 18%, transparent); 
```

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
az108: you can, please use one of the following:

this:

`const copy = structuredClone(issue) as Writeable<ComplianceIssue>;  `

or this whatever you prefer, i personally preffer above i think:

```
const copy = structuredClone(issue);                                                                                                                                                                                                 
  (copy as { language?: string }).language = lang;
```

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -589,6 +593,10 @@ export class JobCreationFormComponent {
     this.basicInfoForm.get('jobDescription')?.setValue(targetContent, { emitEvent: false });
     this.jobDescriptionSignal.set(targetContent);
     this.jobDescriptionEditor()?.forceUpdate(targetContent);
+
+    setTimeout(() => {
+      this.applyHighlights(this.complianceIssues(), newLanguage);
+    }, 50);
```
az108: The 50ms timeout is fragile — it assumes Quill finishes updating within that window, which isn't guaranteed. If forceUpdate doesn't leave the editor ready synchronously, the fix should be there (callback/promise), not a delay here.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1530,12 +1577,17 @@ export class JobCreationFormComponent {
...
         }
         if (attempt === 0) {
           await new Promise(resolve => setTimeout(resolve, 500));
         }
       }
+      const currentLang = this.currentDescriptionLanguage();
+      if (currentLang === lang) {
+        this.applyHighlights(incomingLang, lang);
+      }
+      this.applyHighlights(compliance, lang);
```
az108: Is this intended? applyHighlights(incomingLang, lang) runs conditionally, then applyHighlights(compliance, lang) runs unconditionally right after and overwrites it. Should this just be the filtered call inside the if, with no highlight applied when the editor is showing a different language?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1520,8 +1556,19 @@ export class JobCreationFormComponent {
...
     try {
       // 2) Send the description to the analysis endpoint (persists score on the backend)
-      await firstValueFrom(this.aiApi.analyzeJobDescriptionForCompliance(lang, jobForm));
+      const compliance = await firstValueFrom(this.aiApi.analyzeJobDescriptionForCompliance(lang, jobForm));
       this.lastAnalyzedText = descriptionText;
+      // Keep issues from other languages, but replace all issues for the current language with the latest analysis.
+      const otherLang = lang === 'en' ? 'de' : 'en';
+      const existingLang = this.complianceIssues().filter(issue => issue.language === otherLang);
+
+      const incomingLang = compliance.map(issue => {
```
az108: do we even need this as the server already sets language in AiService.setLanguage(lang), so this client-side patching shouldn't be needed?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1477,7 +1512,8 @@ export class JobCreationFormComponent {
               const saved = await firstValueFrom(this.jobApi.updateJob(jobId, currentData));
               this.lastSavedData.set(saved);
               this.isAnalyzing.set(true);
-              void this.analyzeAndUpdateScore(targetLang);
+              await this.analyzeAndUpdateScore(currentLang);
+              await this.analyzeAndUpdateScore(targetLang);
```
az108: These two calls are independent. Could they run in parallel with Promise.all([...]) to avoid doubling the wait time?

az108 (changes requested):

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
CelineLahnor: Yes I agree

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: Yes true it isn't. I will remove this one  this.applyHighlights(compliance, lang);

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: Good point, then I will add an onComplete callback in forceUpdate to replace this

src/main/resources/prompts/TranslateText.st
ge94zec: Since adding the compliance checker, keeping gender-inclusive instructions in the 
translation prompt creates a direct conflict, the translator flags or refuses to translate non-inclusive/non-compliant terms instead of just translating them, which breaks the flow. Looking ahead, once the gender decoder is also in place, having the translator silently correct language that the decoder is supposed to flag will create consistency issues between the two language versions. The user would see the decoder flagging 
issues in one language but not in the other, because the translator already "fixed" 
them, which defeats the purpose of the decoder and compliance analysis entirely. 

I temporarily reverted the change for now, but maybe we need to discuss this one further more tomorrow.

az108: i understand the concern, i would still not remove the gender inclusive instruction, what might be better would be to modify the prompt to make sure that the translator does not refuse to translate non inclusive terms. Something like prefer inclusive language but make sure the translation is complete.

Cathy0123456789 (approved):
Code LGTM, not tested though

az108 (approved):
Code + Tested Locally 👍

src/main/resources/prompts/TranslateText.st
ge94zec: This might be confusing, since AI prefers strict and clear instructions without much room for interpretation. I’ll think about how to improve the prompts.

PR #2310, opened 2026-04-17
`Development`: Refactor AI toggle, add streaming translation with SSE, and improve score display

### LINKED ISSUE #2219: Relocate Job Description Instructions to Info Icon
### Description

The Job Creation Form currently displays permanent instructional text for the "Job Description" field. While helpful for first-time users, this text remains visible for every subsequent visit, occupying valuable screen real estate and creating unnecessary visual noise.

Experienced users who have already internalized these guidelines find the persistent text redundant. Moving this information into an info icon streamlines the interface for power users while ensuring that the guidance remains available on-demand for those who need a refresher.

Image

### Tasks

- Implement a tooltip or hover-state popover to house the instructional text.

### Acceptance Criteria

_No response_

### LINKED ISSUE #2224: Enable language switching during AI translation and form saving
### Description

Currently, the language switcher is disabled whenever the system is performing an AI translation or saving the form. This restriction prevents users from reviewing other localized versions of the content or adjusting settings while background processes are running.

### Tasks

_No response_

### Acceptance Criteria

- [ ] The language switcher remains interactive and clickable while "Saving" or "Translating" states are active.
- [ ] A progress spinner is displayed within the editor workspace if the user switches to a language that is currently being processed by the AI.
- [ ] No data loss occurs when switching languages during a "Saving" state.

### LINKED ISSUE #2225: Prevent Navigation with Invalid Multilingual Job Descriptions
### Description

Currently, the "Next" button remains active even if a job description field is empty or exceeds the character limit in one of the required languages. This allows users to proceed through the job creation workflow with incomplete data.

### Tasks

- Implement a validation check for the character count in both language input fields.
- Implement a validation check so job descriptions don't equal the template text

### Acceptance Criteria

- [ ] The "Next" button is disabled if the character count in either language field is 0
- [ ] The "Next" button is disabled if the character count in either language field exceeds the maximum limit
- [ ] The "Next" button becomes enabled immediately once both fields meet all validation rules
- [ ] Error messages clearly indicate which language field is failing and why (e.g., "English description too long")

### LINKED ISSUE #2309: Refactor AI toggle, streaming translation, and score display in job creation form
## Description

The job creation form's AI features need several improvements:

### AI Toggle
- The AI enabled/disabled toggle is currently a simple toggle switch buried in the description section. It should be a segmented toggle in the page header (above the stepper) so it's not affected by layout changes when the AI sidebar appears/disappears.
- Add a "?" info button with a dialog explaining what AI enabled/disabled means.

### Streaming Translation
- Language switching is currently blocked during translation, which is a poor UX.
- The translation uses a blocking REST call instead of SSE streaming.
- There's no way to cancel an in-progress translation.
- If the user edits while translation is running, the old translation result could overwrite their changes.

### Translation Robustness
- Translation direction should be clearly visible (EN → DE / DE → EN).
- Both English and German descriptions should be mandatory before proceeding.
- The autosave effect re-triggers on translation state changes, causing translations to cancel and restart.

### Score Display
- The inclusivity score shows 0% before any analysis has run (should show N/A).
- The score doesn't grey out during analysis.
- Score fetch after analysis is unreliable due to `@JsonInclude(NON_EMPTY)` and DB transaction timing.

### Performance
- After AI generation, there's a 5-second autosave delay before translation and analysis start.
- Translation and analysis should run in parallel.
- Autosave debounce is too long (5s → 2s).

az108:
## Summary

Closes #2309

- **Move AI toggle** from description section to page header as a segmented toggle (`AI Enabled` / `AI Disabled`) with an info dialog explaining the modes
- **Replace blocking REST translation** with SSE streaming — supports real-time editor updates, cancellation via AbortController, and allows language switching during translation
- **Show translation direction** in the indicator badge (e.g. `Translating EN → DE`)
- **Enforce both languages** — Next button disabled until both EN and DE descriptions are filled, with a note explaining the requirement
- **Fix autosave effect** — use `untracked()` to prevent translation/generation state changes from re-triggering the autosave timer
- **Optimize post-generation workflow** — after AI generation, immediately save + analyze + translate in parallel (skip 5s autosave delay)
- **Improve score display** — show `N/A` instead of `0%` when no analysis has run; grey out score ring during generation or analysis; retry score fetch after analysis to handle `@JsonInclude(NON_EMPTY)` and DB transaction timing
- **Reduce autosave debounce** from 5s to 2s
- **Clean up dead code** — remove unused translation keys (`languageInfo`, `aiToggle`), unused `TranslateComplianceDTO` import, and `ToggleSwitchComponent`/`ToggleSwitchModule` from the form

## Test plan

- [ ] Verify AI Enabled/Disabled segmented toggle appears in the page header next to the title
- [ ] Verify clicking the "?" button opens the AI info dialog
- [ ] Verify language switching works during active translation (toggle not disabled)
- [ ] Verify translation streams into the editor when viewing the target language tab
- [ ] Verify translation direction indicator shows correctly (e.g. `Translating EN → DE`)
- [ ] Verify editing on the source language tab while translation runs cancels and restarts translation with new content
- [ ] Verify both EN and DE descriptions are required (Next button disabled if either is empty)
- [ ] Verify score shows `N/A` before first analysis, then updates to actual percentage after analysis completes
- [ ] Verify score ring greys out during generation and analysis
- [ ] Verify AI generation immediately triggers save + analysis + translation (no 5s delay)
- [ ] Verify autosave triggers after 2s debounce
- [ ] Test in both English and German UI languages

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -491,7 +537,7 @@ export class JobCreationFormComponent {
    */
   changeDescriptionLanguage(newLang: Language): void {
```
CelineLahnor, line 538: The changeDescriptionLanguage() method clears the debounce timer and only copies the editor HTML into the EN/DE signals. But It never persists that state because onStepChange() also saves only when autoSaveTimer is still set, so typing and then switching languages within the 2s window can leave the latest text unsaved until another edit happens.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1237,33 +1362,31 @@ export class JobCreationFormComponent {
...
-        this.aiScore.set(saved.genderBiasScore);
-      }
-
       this.savingState.set('SAVED');
 
+      // 4) Fire translation (fire-and-forget). Analysis runs once at the end
+      //    of translation after both languages are available — avoids duplicate
+      //    analysis calls that cause score flash issues.
       if (this.aiToggleSignal()) {
-        void this.analyzeAndUpdateScore(currentLang, saved);
```
CelineLahnor, line 1265: Is there a reason you removed this analyzeAndUpdateScore(...) call from the normal autosave path? As long as I understood is without it the score recalculation no longer happens for the common cases where translation is skipped e.g. AI is disabled or both languages are edited manually. In those cases the descriptions are saved, but the sidebar score stays stale (or N/A) because you don't send an analysis request anymore.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -299,6 +321,13 @@ export class JobCreationFormComponent {
   /** Signal that emits when positionDetailsForm status changes */
   positionDetailsChanges = toSignal(this.positionDetailsForm.statusChanges, { initialValue: this.positionDetailsForm.status });
 
+  /** Computed: true when both EN and DE job descriptions have non-empty content */
+  bothDescriptionsFilled = computed(() => {
```
CelineLahnor, line 325: This new validity check reads jobDescriptionEN() / jobDescriptionDE(), but those signals are only synchronized from the editor on autosave or tab switch. So now when the user is typing the missing second description, the visible editor is already filled, but bothDescriptionsFilled() still stays false until the 2-second debounce fires. As a result, the Next button remains disabled even though both descriptions are already present in the UI.

CelineLahnor (commented):
Thank you for the refactor! I left some comments in the Code :)

Cathy0123456789 (commented):
Tested on TS, found 2 small bugs:
- Next can be pressed even if 1 language is only provided
- Saving Changes gets stuck if you type a job description and change the language before the 2 second saving timer is done

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
Cathy0123456789: As the sidebar is collapsed when the AI toggle is off, I think calculating a score might be redundant then currently? This would need to be changed later though

Cathy0123456789 (commented):
Also, when both languages are filled, Next is pressable, but when then removing the visible language completely, Next is disabled for the time during saving, but then becomes enabled again, even though the job description for the language is empty

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
az108: if ai is disabled then so is the sidebar with the score, so therefore this case cant happen here

Issue #1385, opened 2025-11-28
Show validation message when country and postal code do not match


Bofan-Zhu:
### Description

When a postal code is already filled out and the user selects a country that does not match the postal code, the Next button becomes disabled but no error message is shown to explain why.

<img width="2048" height="393" alt="Image" src="https://github.com/user-attachments/assets/fba55032-196a-47de-b90e-2fbb77fbc153" />

This results in poor UX: users do not understand why they cannot proceed, and it appears as if the form is broken.

We should display a clear and accessible validation message when the postal code is invalid for the chosen country. The message should appear near the postal code field and optionally also appear in the sticky footer.

### Tasks

1. Display a clear validation message under the postal code input, for example: “The postal code does not match the selected country.”

### Acceptance Criteria

- [ ] When user selects a country that does not match the postal code: A validation error appears below the postal code field and the error text clearly explains the mismatch

PR #2314, opened 2026-04-18
`Bugfix`: Fix correct deadline/start date validation in job creation form

### LINKED ISSUE #1487: Prevent invalid job dates where start date is earlier than application deadline
### Description

When a professor creates a job listing they can set an application deadline and a start date. The system currently allows the start date to be earlier than the application deadline. This creates invalid job configurations and can confuse applicants and professors. The form should enforce a logical date order and guide the user to correct mistakes.

The expected behaviour is that the start date must always be the same day or later than the application deadline.

### Tasks

1. Restrict the date selection to only dates after the application deadline for start date or add validation 

### Acceptance Criteria

- [ ] Start date must be after application date

CelineLahnor:
#### General

#### Client

### Motivation and Context

Closes #1487 

### Description

This PR prevents invalid date combinations in the professor job creation form by enforcing that the start date must be the same day as or later than the application deadline.

**Changes**
- added client-side validation so startDate must be the same day as or later than applicationDeadline
- disabled invalid dates in both date pickers depending on which date is selected first
- highlighted the reference date in the datepicker with a tooltip
- extracted shared local date parsing logic
- moved datepicker styling to Tailwind
- added some small client tests for this behavior 

### Steps for Testing

1. Log in as Professor
2. Create a new position draft
3. Select a Start Date
4. Open the Deadline Date Picker and verify that the User can not pick a Deadline after the Start Date
5. Verify the Start Date is highlighted within the Date Picker
6. After selecting a Deadline verify this is also highlighted in the Date Picker for the Start Date
7. Verify that the User can not pick a Start Date before the Deadline

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

### Screenshots

https://github.com/user-attachments/assets/46dadbce-7734-44d7-a724-5d1afd3eccd7

I also updated the job creation form so the start date picker appears on the left and the application deadline picker on the right, matching the intended UI order.
Bildschirmfoto 2026-04-19 um 16 23 26

#### Client

| job-creation-form.component.ts | 0.00% | 1208 | 72 | 6.0 |
| datepicker.component.ts | 100.00% | 167 | 22 | 13.2 |
| date-time.util.ts | 45.16% | 57 | ? | ? |
| date-order-validator.ts | 0.00% | 18 | ? | ? |

_Last updated: 2026-05-17 08:15:26 UTC_


src/main/webapp/app/interview/interview-process-detail/slots-section/slot-creation-form/slot-creation-form.component.ts
```diff
@@ -24,6 +24,28 @@ import { HttpErrorResponse } from '@angular/common/http';
...
+  '[&_.p-datepicker-month:not(.p-disabled):hover]:text-text-on-primary',
+  '[&_.p-datepicker-year:not(.p-disabled):hover]:bg-primary-default',
+  '[&_.p-datepicker-year:not(.p-disabled):hover]:text-text-on-primary',
+];
+
+const CALENDAR_SELECTION_CLASSES = [
+  '[&_.p-highlight]:bg-primary-default',
+  '[&_.p-highlight]:text-text-on-primary',
+  '[&_.p-highlight]:border-primary-default',
+];
```
az108: can you make these readonly please?

src/main/webapp/app/shared/components/atoms/datepicker/datepicker.component.html
```diff
@@ -34,15 +34,32 @@
...
     [dateFormat]="dateFormat()"
     iconDisplay="button"
-    inputId="startDate"
+    [inputId]="resolvedInputId()"
     showIcon="true"
-    [inputStyleClass]="'w-full'"
+    [inputStyleClass]="inputStyleClass"
   >
+    <ng-template pTemplate="date" let-date>
+      <span
```
az108, line 46: isHighlightedDate(date) is called 3 times in this template block, method calls in templates re-evaluate on every change detection cycle. Per our client guidelines: "Keine Methodenaufrufe in Templates   
  (ausser Signal-Getter)".                                                                                                                                                                                     
                                                                                                                                                                                                             
  Since this runs inside pTemplate="date" (once per calendar cell), consider at minimum deduplicating the calls. You could introduce a local variable or simplify the bindings — e.g. [pTooltip] doesn't need  
  to re-check isHighlightedDate() since [tooltipDisabled] already gates it:                                                                                                                                  
                                                                                                                                                                                                               
  <span                                                                                                                                                                                                      
    class="inline-flex size-full items-center justify-center"
    [class.datepicker-reference-day]="isHighlightedDate(date)"                                                                                                                                                 
    [pTooltip]="shouldTranslate() ? (highlightedDateLabel() ?? '' | translate) : (highlightedDateLabel() ?? '')"
    tooltipPosition="top"                                                                                                                                                                                      
    [tooltipDisabled]="!isHighlightedDate(date) || !highlightedDateLabel()"                                                                                                                                    
  >                                                                                                                                                                                                            
    {{ date.day }}                                                                                                                                                                                             
  </span>                                                                                                                                                                                                      
                                                                                                                                                                                                             
  This reduces it from 3 calls to 2 and removes the nested ternary.

az108 (changes requested):

src/main/webapp/i18n/de/global.json
```diff
@@ -344,7 +344,8 @@
       "status": {
         "SAVING": "Änderungen werden gespeichert...",
         "SAVED": "Änderungen gespeichert",
-        "FAILED": "Speichern fehlgeschlagen"
+        "FAILED": "Speichern fehlgeschlagen",
+        "VALIDATION_BLOCKED": "Das Startdatum muss am oder nach der Bewerbungsfrist liegen"
```
Cathy0123456789: "an", not "am" I guess

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1499,6 +1524,12 @@ export class JobCreationFormComponent {
    * Triggers translation after successful save if content changed.
    */
   private async performAutoSave(): Promise<void> {
+    if (this.hasInvalidDateOrder()) {
+      this.savingState.set(SavingStates.VALIDATION_BLOCKED);
+      this.positionDetailsForm.markAllAsTouched();
+      return;
+    }
```
Cathy0123456789: Do we even need this? When I enter something invalid, it anyway does not keep the value but just discards it?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
Cathy0123456789, line 1: I would not exchange the order of the dates, since the application date should be before the start date, it would also make more sense to keep it on the left side?

Cathy0123456789 (changes requested):
Small comments on the code, also I tested locally and realized something. It might not be relevant though just now since prod was wiped but still worth mentioning perhaps:
When having set "invalid" start date and application deadlines before switching to this branch, the dates are not shown in the datepicker field but in the calendar they are:
Image

Also, in general the behavior is: When entering an invalid date directly in the field, it just discards it and does not save it. I am not sure if this is good UX, but when I think of the postcode field for example, we show an error message under the field (at least it's intended #1385) rather than just erasing it. Maybe this could make sense here as well?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
CelineLahnor: I think selecting the start date first and then deciding the application deadline based on that is a more natural and common workflow. If you have to choose the deadline first, you already need to have the start date in mind anyway to set a reasonable deadline. Because usually, people don’t plan the start date around the deadline, they set the deadline based on the intended start date. I also discussed this with @az108 before

CelineLahnor: I changed the order to its original state now

src/main/webapp/app/shared/components/atoms/datepicker/datepicker.component.ts
```diff
@@ -188,4 +258,13 @@ export class DatePickerComponent {
       this.selectedDateChange.emit(undefined);
     }
   }
+
+  isHighlightedDate(date: { day: number; month: number; year: number }): boolean {
```
az108: please no direct method call in html except its a click or similar. please turn this into a computed signal or so

az108 (changes requested):

az108 (approved):

PR #2334, opened 2026-04-20
`Bugfix`: Title compliance error not shown

### LINKED ISSUE #2321: Title compliance error not shown
### Description

After merging, the title compliance error no longer appears for the string input. Additionally, the current sequence is inefficient, and highlights are lost after page reloads or when switching pages. Persisting highlights across page transitions is essential.
It would also be beneficial to adjust the workflow. First analyze the content in its original language, then perform the translation along with the target-language analysis. This change should improve overall performance and responsiveness.

### Tasks

_No response_

### Acceptance Criteria

_No response_

ge94zec:
❗ DO NOT MERGE BEFORE https://github.com/ls1intum/tum-apply/pull/2358  ❗

#### General

#### Server

#### Client

### Motivation and Context

Closes: https://github.com/ls1intum/tum-apply/issues/2321

### Description

Reordered AI calls in client interaction, analysis and highlighting of the current language now run first, while translation and the target language analysis are fire-and-forget in the background. A `LEFT JOIN` query was added in `JobRepository` so compliance issues are eagerly fetched when `getJobId` is called. Additionally, a new effect ensures that highlights are reapplied after page reloads, tab switches, and language changes.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply as Prof 
2. Nav to Create new Position 
3. Verify compliance issues are present in jobId
4. Verify highlights are present across page reloads/ switches 

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [ ] Test 1

### Screenshots

Bildschirmfoto 2026-04-28 um 12 50 11

#### Client

| job-creation-form.component.ts | 68.74% | 1189 | 71 | 6.0 |
| editor.component.ts | 78.53% | 353 | 24 | 6.8 |

#### Server

| AiService.java | 7.48% | 272 |
| JobRepository.java | 0.00% | 203 |
| JobService.java | 73.87% | 332 |

_Last updated: 2026-05-12 16:09:20 UTC_


src/main/java/de/tum/cit/aet/job/service/JobService.java
```diff
@@ -477,9 +478,19 @@ public void updateAiAnalysis(UUID jobId, int score, List<ComplianceIssue> compli
             return;
         }
 
-        Job job = jobRepository.findById(jobId).orElseThrow(() -> EntityNotFoundException.forId("Job", jobId));
+        Job job = jobRepository.findByIdWithCompliance(jobId).orElseThrow(() -> EntityNotFoundException.forId("Job", jobId));
+        String incomingLang = complianceAnalysis.isEmpty() ? null : complianceAnalysis.get(0).getLanguage();
```
CelineLahnor: Can you pass the language explicitly to updateAiAnalysis(...) and always clear the old issues for that language before saving the new result. Because the current approach works only if the analysis returns at least one issue. If the AI finds no issues, the list is empty, incomingLang becomes null, and the old issues for that language stay in the database. Then users can still see old description warnings even though they already fixed them.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1640,6 +1672,15 @@ export class JobCreationFormComponent {
         this.translationTargetLang.set(undefined);
         this.translationAbortController = undefined;
       }
+      // Clear only if this is still the same request.
+      // If a newer one exists, keep it to avoid breaking duplicate checks.
+      if (
```
CelineLahnor: Here the block reads activeTranslationRequest without checking if it still exists but cancelTranslation() can set it to undefined before. Can you please add a guard?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -853,15 +858,29 @@ export class JobCreationFormComponent {
...
+  highlightsEffect = effect(() => {
+    const editor = this.jobDescriptionEditor();
+    const lang = this.currentDescriptionLanguage();
+    const issues = this.complianceIssues();
+    const filter = this.activeComplianceFilter();
+    if (!editor) return;
+    if (untracked(() => this.isGeneratingDraft() || (this.isTranslating() && this.translationTargetLang() === lang))) return;
+    const filtered = filter ? issues.filter(i => i.category === filter) : issues;
+
+    const content = untracked(() => (lang === 'en' ? this.jobDescriptionEN() : this.jobDescriptionDE()));
```
CelineLahnor: I would not set the editor content again at all. Just keep the current editor text and only reload the highlight formatting. If we set the editor content again inside highlightsEffect, we risk replacing the text the user is currently typing with older stored content. That can happen because jobDescriptionEN / jobDescriptionDE may still be behind the live editor state.

CelineLahnor (changes requested):
LGTM! I left a few comments in the code please check it out :)

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: l will avoid accessing properties on a cleared request by checking if activeRequest is the same object.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: I agree with the concern. The reason for forceUpdate here was the reload case. The concrete issue I saw is, after reloading and opening the draft, the persisted compliance issues are loaded, but the highlights are not visible in the editor. If I then switch the translation language, the highlights appear. That happens because the language switch calls forceUpdate, so Quill is explicitly loaded with the stored content before applyHighlights runs. Without that explicit content update during reload, highlightTexts can run against an editor state that is not ready with the form content, so it does not apply anything.

ge94zec: Maybe you or @az108 have an idea how to do it diffrently?

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
CelineLahnor: This change does not implement the intended guard. In the finally block, activeRequest is recreated and then assigned back into this.activeTranslationRequest before the comparison, so the condition is always true and the code will clear the field unconditionally. The safe pattern is to capture the original request object before the await, for example when starting the translation, and in finally only do:

if (this.activeTranslationRequest === activeRequest) {
  this.activeTranslationRequest = undefined;
}

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1640,6 +1672,13 @@ export class JobCreationFormComponent {
         this.translationTargetLang.set(undefined);
         this.translationAbortController = undefined;
       }
+      // Clear only if this is still the same request.
```
CelineLahnor, line 1706: Change this added block to:   
```
// Clear only if this exact request is still the active one.
  if (this.activeTranslationRequest === activeRequest) {
    this.activeTranslationRequest = undefined;
  }
```

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1551,14 +1576,21 @@ export class JobCreationFormComponent {
...
+
     // 1) Skip if the text hasn't changed since the last translation
     const lastBaseline = currentLang === 'en' ? this.lastTranslatedEN() : this.lastTranslatedDE();
     if (text === lastBaseline) return;
 
     // 2) Cancel any active translation and set up fresh state
     this.cancelTranslation();
-    const targetLang: Language = currentLang === 'en' ? 'de' : 'en';
     const abortController = new AbortController();
+    this.activeTranslationRequest = { sourceLang: currentLang, sourceText: text, targetLang };
```
CelineLahnor: And change this to:
```
const activeRequest = { sourceLang: currentLang, sourceText: text, targetLang };
this.activeTranslationRequest = activeRequest;
```

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
CelineLahnor: I understand your problem and I would suggest to remove the forceUpdate from highlightsEffect, and make EditorComponent.highlightTexts() reapply after mount or content sync instead.

so something like this here: 
```
highlightsEffect = effect(() => {
  const editor = this.jobDescriptionEditor();
  const lang = this.currentDescriptionLanguage();
  const issues = this.complianceIssues();
  const filter = this.activeComplianceFilter();

  if (!editor) return;
  if (untracked(() => this.isGeneratingDraft() || (this.isTranslating() && this.translationTargetLang() === lang))) return;

  const filtered = filter ? issues.filter(i => i.category === filter) : issues;
  this.applyHighlights(filtered, lang);
});

```
And then solve the reload problem separately by making highlight application resilient when Quill is not ready yet so I would update the **editor.component.ts:** 
```
private pendingHighlights = signal<{ text: string; color: string; bg: string }[]>([]);

public highlightTexts(highlights: { text: string; color: string; bg: string }[]): void {
  this.pendingHighlights.set(highlights);
  this.applyPendingHighlights();
}

private reapplyHighlightsEffect = effect(() => {
  this.quillEditorComponent();
  this.htmlValue();
  this.pendingHighlights();
  requestAnimationFrame(() => this.applyPendingHighlights());
});

private applyPendingHighlights(): void {
  const editor = this.quillEditorComponent()?.quillEditor;
  if (!editor) return;

  const highlights = this.pendingHighlights();
  editor.formatText(0, editor.getLength(), 'background', false);
  editor.formatText(0, editor.getLength(), 'customHighlight', false);

  const fullText = editor.getText().toLowerCase();
  for (const { text, color, bg } of highlights) {
    let startIndex = 0;
    const searchText = text.toLowerCase();
    while (startIndex < fullText.length) {
      const index = fullText.indexOf(searchText, startIndex);
      if (index === -1) break;
      editor.formatText(index, text.length, 'customHighlight', { color, bg });
      startIndex = index + text.length;
    }
  }
}
```

CelineLahnor (changes requested):
I gave you some code suggestions feel free to let me know ur opinion :)

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
CelineLahnor: I left a code suggestion below :)

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
ge94zec: I mostly implemented your suggestion, forceUpdate is now removed from the highlightsEffect, and reload resilience is handled in the EditorComponent via pendingHighlights and an effect. There are two small deviations from your snippet that I will explain:

1.  I dropped `htmlValue()` as an effect dependency. With htmlValue() in there, the effect re-fires on every keystroke, which means re-running formatText over the whole document on every typed character. Other effects already handle this, so the dependency isn’t needed
2.  I added a small retry inside `applyPendingHighlights`. The issue I ran into was that after reloading a draft, the highlights weren’t visible at first. They only showed up once something else triggered an update, for example when switching the language.` applyPendingHighlights` now retries on the next frame if Quill isn’t ready yet. It only does this while there are still pending highlights, so it won’t loop forever when the editor gets destroyed.

CelineLahnor (approved):
This is fine for me thank you !

az108 (approved):

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
```diff
@@ -124,6 +124,8 @@ export class EditorComponent extends BaseInputDirective<string> {
   openAnalysisDialog = output<GenderBiasAnalysisResponse>();
   quillEditorComponent = viewChild(QuillEditorComponent);
   highlightHovered = output<{ text: string; x: number; y: number } | undefined>();
+  highlights = input<{ text: string; category: ComplianceIssueCategoryEnum }[]>([]);
```
az108: this seems to be unused?

src/main/java/de/tum/cit/aet/job/repository/JobRepository.java
```diff
@@ -307,4 +308,14 @@ ORDER BY CONCAT(p.firstName, ' ', p.lastName) ASC
...
     @Query("SELECT DISTINCT j.image.imageId FROM Job j WHERE j.image.imageId IN :imageIds")
     Set<UUID> findInUseImageIds(@Param("imageIds") List<UUID> imageIds);
+
+    /**
+     * Finds a job by id, eagerly fetching compliance issues
+     *
+     * @param jobId the job id
+     * @return the job with relations loaded, or empty if not found
+     */
+    @EntityGraph(attributePaths = { "complianceIssues", "supervisingProfessor", "researchGroup", "image" })
```
az108: The two callers (assertCanManageJob and updateAiAnalysis) only need researchGroup and complianceIssues respectively — image and supervisingProfessor aren't read by either, so this over-fetches on every call. Can we trim the graph to { "complianceIssues", "researchGroup" }?

src/main/java/de/tum/cit/aet/job/service/JobService.java
```diff
@@ -473,15 +474,25 @@ public void updateJobDescriptionLanguage(String jobId, String toLang, String tra
...
-        Job job = jobRepository.findById(jobId).orElseThrow(() -> EntityNotFoundException.forId("Job", jobId));
+        Job job = jobRepository.findByIdWithCompliance(jobId).orElseThrow(() -> EntityNotFoundException.forId("Job", jobId));
+
+        // Keep issues from the other language, add new ones for target language
+        List<ComplianceIssue> issuesToSave = new ArrayList<>();
+        for (ComplianceIssue existingLang : job.getComplianceIssues()) {
+            if (!Objects.equals(existingLang.getLanguage(), lang)) {
+                issuesToSave.add(existingLang);
+            }
+        }
```
az108: existingLang reads as a language but it's actually a ComplianceIssue. Either rename to existing/issue, or use a stream:      
                                                                                                                                                                                                                          List<ComplianceIssue> issuesToSave = job.getComplianceIssues().stream()                                                                                                                                                            
.filter(issue -> !Objects.equals(issue.getLanguage(), lang))                                                                                                                                                                   
.collect(Collectors.toCollection(ArrayList::new));                                                                                                                                                                             
issuesToSave.addAll(complianceAnalysis);

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -903,14 +908,27 @@ export class JobCreationFormComponent {
...
    */
   onComplianceFilterChange(category: string | undefined): void {
     this.activeComplianceFilter.set(category);
+  }
+
+  /**
+   * Handles highlights after reload, page switches, language switches or new analysis.
+   * Skips while AI is actively generating new draft or translating.
+   */
+  highlightsEffect = effect(() => {
```
az108: highlightsEffect isn't used in the template — should be private

az108 (changes requested):

Cathy0123456789 (changes requested):
I think doing the translation after compliance doesn't make sense, since it can run in parallel

src/main/java/de/tum/cit/aet/job/repository/JobRepository.java
Cathy0123456789: Considering the changes made to the tests, maybe we should check again if it would make sense to include the images? @az108

az108 (approved):

PR #2338, opened 2026-04-21
`Development`: Add AI feature disable switch 

### LINKED ISSUE #2337: Add AI feature kill switch (Fehlerstrategie) for system-wide AI disable
## Description

Implement a system-wide mechanism to quickly disable all LLM-dependent AI features without requiring redeployment. This is needed as a safety net in case of:
- Expired or broken AI API keys
- AI provider service outages
- Unexpected billing spikes
- Any other operational emergency requiring AI to be turned off

### Manual Toggle (Admin)
- [ ] Admin-only REST endpoint to enable/disable AI features system-wide
- [ ] State persisted in DB (`system_settings` table) to survive restarts
- [ ] Public status endpoint for frontend to check AI availability
- [ ] Admin UI toggle in the admin dashboard

### Automatic Circuit Breaker
- [ ] Track consecutive LLM call failures
- [ ] Auto-disable AI after N consecutive failures (default: 5)
- [ ] Auto-recover after cooldown period (default: 60 seconds)
- [ ] Admin endpoint to manually reset circuit breaker

### Backend Guards
- [ ] All AI REST endpoints return 503 when AI is disabled
- [ ] Gender bias analysis (word-list based, no LLM) continues working

### Frontend Adaptation
- [ ] Frontend service to check system-wide AI status
- [ ] Disable AI generation, translation, extraction, and analysis when AI is off
- [ ] Show appropriate messages to users when AI is unavailable
- [ ] Handle 503 responses gracefully

### Tests
- [ ] Unit tests for AiFeatureToggleService
- [ ] Integration tests for toggle endpoints
- [ ] Frontend tests for AiFeatureStatusService

## Non-LLM Features (should keep working)
- Gender bias scoring (word-list matching)
- Compliance score calculation (math only)
- User AI consent settings

az108:
#### General

#### Server

- [x] I added multiple server tests (JUnit) related to the features (with a high test coverage), while following the [server test guidelines](https://ls1intum.github.io/tum-apply/developer/server-guidelines/server-tests).

#### Client

### Motivation and Context

Resolves #2337

In production, if the AI API key expires, the provider has an outage, or any other operational emergency occurs, there is currently no way to disable AI features without redeploying. This PR adds a **Fehlerstrategie** (error strategy) — a dual mechanism to disable AI system-wide:

1. **Manual admin toggle** — persisted in DB, instant effect, no restart needed
2. **Automatic circuit breaker** — trips after 5 consecutive LLM failures, auto-recovers after 60s

Non-LLM features (gender bias word-list scoring, compliance score math) continue working when AI is disabled.

#### Server
- **`SystemSetting`** entity + repository — generic key-value table for system-wide config
- **`AiFeatureToggleService`** — core service managing the manual toggle (DB-persisted) and in-memory circuit breaker (5 failures threshold, 2 hour cooldown with half-open probe)
- **`AiFeatureToggleResource`** — 3 endpoints:
  - `GET /api/ai/feature-toggle/status` (authenticated) — returns `{aiEnabled, manuallyDisabled, circuitBreakerOpen}`
  - `PUT /api/ai/feature-toggle/toggle?enabled=bool` (admin) — manual enable/disable
  - `POST /api/ai/feature-toggle/reset-circuit-breaker` (admin) — reset after fixing root cause
- **`AiResource`** — all 4 LLM endpoints now call `ensureAiAvailable()` → returns 503 when disabled
- **`AiService`** — all LLM call sites now report `recordSuccess()`/`recordFailure()` to the circuit breaker
- **`AiServiceUnavailableException`** → maps to HTTP 503 via `GlobalExceptionHandler`
- **Liquibase** changelog `00000000000034` creates `system_settings` table with default `ai.enabled=true`

#### Client
- **`AiFeatureStatusService`** — singleton service fetching system AI status, exposes `aiSystemEnabled` signal
- **`AiStreamingService`** — detects 503 and calls `markUnavailable()`
- **`JobCreationFormComponent`** — guards generate, translate, and analyze with `aiSystemEnabled()`
- **`AiExtractionBoxComponent`** — shows toast when AI is disabled (consistent with professor flow)

### Steps for Testing

1. Log in to TUMApply as an admin
2. Verify AI features work (generate job description, translate, extract PDF, compliance analysis)
3. Navigate to **System Settings** page (`/system-settings`)
4. Toggle AI features **off** using the toggle switch
5. Verify all AI actions show a toast ("AI Unavailable") instead of functioning — test both professor (job creation) and applicant (PDF extraction) flows
6. Verify gender bias analysis (non-LLM) still works
7. Toggle AI features **on** again from System Settings
8. Verify AI features work again
9. To test circuit breaker: configure an invalid API key, trigger 5+ AI calls, verify the circuit breaker tag on System Settings shows "Open" and the reset button appears

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

#### Client

| admin-system-settings.component.ts | 100.00% | 82 | 27 | 32.9 |
| job-creation-form.component.ts | 0.00% | 1162 | 126 | 10.8 |
| ai-feature-status.service.ts | 0.00% | 32 | ? | ? |
| ai-streaming.service.ts | 0.00% | 86 | 18 | 20.9 |
| ai-extraction-box.component.ts | 0.00% | 129 | ? | ? |

#### Server

| SystemSetting.java | 0.00% | 21 |
| AiFeatureStatusDTO.java | 100.00% | 10 |
| SystemSettingRepository.java | 0.00% | 6 |
| AiFeatureToggleService.java | 72.34% | 83 |
| AiService.java | 7.48% | 272 |
| AiFeatureToggleResource.java | 100.00% | 38 |
| AiResource.java | 43.48% | 87 |

_Last updated: 2026-05-01 22:01:19 UTC_


src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,136 @@
...
+ * </ul>
+ *
+ * AI is available only when both conditions are met: manually enabled AND circuit closed.
+ */
+@Service
+public class AiFeatureToggleService {
+
+    private static final String SETTING_KEY = "ai.enabled";
+    private static final int FAILURE_THRESHOLD = 5;
+    private static final long COOLDOWN_SECONDS = 60;
```
Cathy0123456789: Is 60s cooldown not too short?

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private final SystemSettingRepository systemSettingRepository;
+
+    private final AtomicBoolean manuallyEnabled = new AtomicBoolean(true);
+    private final AtomicInteger consecutiveFailures = new AtomicInteger(0);
+    private final AtomicLong circuitOpenedAt = new AtomicLong(0);
```
Cathy0123456789: Do we need Atomic classes?

src/main/java/de/tum/cit/aet/ai/web/AiResource.java
```diff
@@ -115,6 +128,7 @@ public ResponseEntity<List<ComplianceIssue>> analyzeJobDescriptionForCompliance(
         @RequestParam("lang") String descriptionLanguage,
         @RequestParam(defaultValue = "en") String userLanguage
     ) {
+        ensureAiAvailable();
```
CelineLahnor: Can we separate the non-LLM compliance work from the LLM-based part in the analyze-job-description flow, instead of returning 503 for the whole endpoint? Right now, when AI is disabled, the guard in AiResource stops the request before analyzeCurrentJobDescription() runs, so the server also skips the local GenderBiasAnalysisService logic. I would keep the rule-based scoring and score calculation available even when the LLM-based analysis is unavailable.

src/main/webapp/app/admin/system-settings/admin-system-settings.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  templateUrl: './admin-system-settings.component.html',
+})
+export class AdminSystemSettingsComponent {
+  /** Whether the initial status is being loaded. */
+  readonly isLoading = signal(false);
+
+  /** Whether a toggle or reset action is in progress. */
+  readonly isUpdating = signal(false);
+
+  /** Current AI feature status from the backend. */
```
Cathy0123456789: backend

src/main/webapp/app/admin/system-settings/admin-system-settings.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  readonly manuallyDisabled = computed(() => this.aiStatus()?.manuallyDisabled ?? false);
+
+  private readonly aiFeatureToggleApi = inject(AiFeatureToggleResourceApi);
+  private readonly toastService = inject(ToastService);
+
+  constructor() {
+    void this.loadStatus();
+  }
+
+  /** Fetches the current AI feature status from the backend. */
```
Cathy0123456789: backend

src/main/webapp/app/service/ai-feature-status.service.ts
```diff
@@ -0,0 +1,62 @@
...
+  readonly loaded = signal<boolean>(false);
+
+  private aiFeatureToggleApi = inject(AiFeatureToggleResourceApi);
+
+  constructor() {
+    void this.refresh();
+  }
+
+  /**
+   * Fetches the current AI feature status from the backend.
```
Cathy0123456789: backend

src/main/java/de/tum/cit/aet/ai/web/AiFeatureToggleResource.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final AiFeatureToggleService aiFeatureToggleService;
+
+    public AiFeatureToggleResource(AiFeatureToggleService aiFeatureToggleService) {
+        this.aiFeatureToggleService = aiFeatureToggleService;
+    }
+
+    /**
+     * Returns the current system-wide AI feature status.
+     * Available to any authenticated user so the frontend can adapt the UI.
```
Cathy0123456789: frontend

src/main/java/de/tum/cit/aet/ai/web/AiResource.java
```diff
@@ -27,9 +29,17 @@
...
 
-    public AiResource(AiService aiService) {
+    public AiResource(AiService aiService, AiFeatureToggleService aiFeatureToggleService) {
         this.aiService = aiService;
+        this.aiFeatureToggleService = aiFeatureToggleService;
+    }
+
+    private void ensureAiAvailable() {
+        if (!aiFeatureToggleService.isAiAvailable()) {
+            throw new AiServiceUnavailableException("AI features are currently disabled");
```
Cathy0123456789: I think this will pollute the server log. Can we remove the exception?

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,136 @@
...
+     * to allow a probe request.
+     *
+     * @return {@code true} if the circuit breaker is currently open
+     */
+    public boolean isCircuitBreakerOpen() {
+        long openedAt = circuitOpenedAt.get();
+        if (openedAt == 0) {
+            return false;
+        }
+        if (System.currentTimeMillis() - openedAt > COOLDOWN_SECONDS * 1000) {
```
CelineLahnor: Can we model the circuit breaker with explicit CLOSED / OPEN / HALF_OPEN states and allow exactly one half-open probe instead of relying only on timestamps? Right now, after the first cooldown, isCircuitBreakerOpen() starts returning false once circuitOpenedAt is older than 60s, but failed probes do not actually re-open the breaker because recordFailure() only opens it when circuitOpenedAt == 0. That means the breaker can become effectively permissive after the first cooldown unless a successful call resets it first.

src/main/webapp/app/service/ai-feature-status.service.ts
```diff
@@ -0,0 +1,62 @@
...
+   * Components can call this after receiving a 503 to update UI state.
+   */
+  async refresh(): Promise<void> {
+    try {
+      const status = await firstValueFrom(this.aiFeatureToggleApi.getAiStatus());
+      this.aiSystemEnabled.set(status.aiEnabled ?? false);
+      this.manuallyDisabled.set(status.manuallyDisabled ?? false);
+      this.circuitBreakerOpen.set(status.circuitBreakerOpen ?? false);
+    } catch {
```
CelineLahnor: Can we avoid treating every failure when fetching the AI feature status as “AI unavailable” on the client? refresh() is called immediately in the service constructor, but this catch block sets aiSystemEnabled to false for any error. That means a temporary network issue, an authentication problem during startup, or another unrelated server error would all disable AI in the UI even if the system toggle is still on. It may be safer to only mark AI unavailable after a confirmed 503, and otherwise keep the previous state or expose a loading or unknown state.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -873,6 +878,10 @@ export class JobCreationFormComponent {
    * After generation completes, the final content is force-updated to ensure correctness.
    */
   async generateJobApplicationDraft(): Promise<void> {
+    if (!this.aiSystemEnabled()) {
+      this.toastService.showErrorKey('ai.featureToggle.systemDisabled');
```
CelineLahnor: Would it make sense to vary this message more based on whether AI was manually disabled or the circuit breaker is open? Right now the same ai.featureToggle.systemDisabled toast is used for all unavailable states, but its text says AI was disabled by the system administrator, which may be misleading when the unavailability is caused by the circuit breaker instead.

CelineLahnor (changes requested):
LGTM! I left some suggestions in the code :)

Cathy0123456789 (changes requested):
Thank you, small comments on the code

src/main/java/de/tum/cit/aet/ai/domain/SystemSetting.java
```diff
@@ -0,0 +1,32 @@
...
+    @Column(name = "setting_key", nullable = false)
+    private String key;
+
+    @Column(name = "setting_value", length = 1000)
+    private String value;
+
+    public SystemSetting(String key, String value) {
+        this.key = key;
+        this.value = value;
+    }
```
Cathy0123456789: Can we replace this with @AllArgsConstructor?

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Record a successful LLM call. Closes the circuit breaker and resets failure count.
+     */
+    public void recordSuccess() {
+        synchronized (circuitLock) {
+            consecutiveFailures = 0;
+            circuitState = CircuitState.CLOSED;
+            openedAt = 0;
+        }
+    }
```
Cathy0123456789, line 110: This method looks same like resetCircuitBreaker, can we combine them maybe?

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,159 @@
...
+                openedAt = System.currentTimeMillis();
+                return;
+            }
+            consecutiveFailures++;
+            if (circuitState == CircuitState.CLOSED && consecutiveFailures >= FAILURE_THRESHOLD) {
+                circuitState = CircuitState.OPEN;
+                openedAt = System.currentTimeMillis();
+            }
+        }
+    }
```
Cathy0123456789, line 126: We could shorten the method maybe:
```
public void recordFailure() {
    synchronized (circuitLock) {
        if (circuitState == CircuitState.HALF_OPEN || 
           (circuitState == CircuitState.CLOSED && ++consecutiveFailures >= FAILURE_THRESHOLD)) {
            circuitState = CircuitState.OPEN;
            openedAt = System.currentTimeMillis();
        }
    }
}
```

src/main/java/de/tum/cit/aet/ai/web/AiResource.java
```diff
@@ -115,6 +131,9 @@ public ResponseEntity<List<ComplianceIssue>> analyzeJobDescriptionForCompliance(
         @RequestParam("lang") String descriptionLanguage,
         @RequestParam(defaultValue = "en") String userLanguage
     ) {
+        // Intentionally NOT calling ensureAiAvailable(): the rule-based gender bias analysis
+        // and score computation must remain available even when LLM-backed legal analysis is off.
+        // The service skips the LLM call internally when AI is disabled.
```
Cathy0123456789: Can we maybe shorten the comment like:
`// Service skips LLM calls internally when AI is disabled, rule-based gender bias analysis and score computation remain enabled`

Cathy0123456789 (changes requested):
Thank you, small comments still

Cathy0123456789 (changes requested):
Currently when the circuit breaker is open, it is unclear when it was opened and what the cooldown period is. Maybe we can add that?
Image

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
CelineLahnor: Since you changed it to 300 now can you please update the PR description because there it still says the breaker should auto-recover after 60 seconds.

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -0,0 +1,153 @@
...
+            openedAt = 0;
+        }
+    }
+
+    /**
+     * Returns a snapshot of the current AI feature status.
+     *
+     * @return the current AI feature status including manual toggle and circuit breaker state
+     */
+    public AiFeatureStatusDTO getStatus() {
```
CelineLahnor, line 155: getStatus() checks circuitState directly, so it can keep reporting the breaker as open even after the cooldown expired. The OPEN -> HALF_OPEN transition only happens in isCircuitBreakerOpen(), and because the client blocks requests based on the status response, the probe request may never run. I’d suggest using isCircuitBreakerOpen() when building the status so the breaker can recover automatically as intended.

CelineLahnor (changes requested):
Thank you. Code looks already rly good I just noticed one small thing

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
```diff
@@ -141,13 +141,22 @@ public void resetCircuitBreaker() {
...
         synchronized (circuitLock) {
-            breakerOpen = circuitState != CircuitState.CLOSED;
+            openedAtSnapshot = openedAt;
+            boolean cooldownElapsed = circuitState == CircuitState.OPEN && System.currentTimeMillis() - openedAt > COOLDOWN_SECONDS * 1000;
+            breakerOpen = circuitState != CircuitState.CLOSED && !cooldownElapsed;
         }
-        return new AiFeatureStatusDTO(manuallyEnabled && !breakerOpen, !manuallyEnabled, breakerOpen);
+        return new AiFeatureStatusDTO(manuallyEnabled && !breakerOpen, !manuallyEnabled, breakerOpen, COOLDOWN_SECONDS, openedAtSnapshot);
     }
 }
```
Cathy0123456789, line 165: Instead of adding a local variable openedAtSnapshot, would it make sense to just create the DTO while holding the lock?

Cathy0123456789 (changes requested):
Small question

src/main/java/de/tum/cit/aet/ai/service/AiFeatureToggleService.java
az108: Could do, but I'd rather keep the lock scoped to just the circuit state it protects — manuallyEnabled is a separate volatile field and pulling it inside circuitLock mixes concerns. The snapshot pattern also keeps the critical section minimal, which is the conservative default. Happy to change it if you'd prefer the tighter version though.

Cathy0123456789: Makes sense then to keep it this way, thank you 👍🏼

CelineLahnor (approved):
Code looks fine from my side now. Thank you :)

Cathy0123456789 (approved):
Thank you, tested locally and code LGTM as well

PR #2349, opened 2026-04-23
`Development`: Add new prompt categories

### LINKED ISSUE #2344: Extend prompt with stricter rules
### Description

There is still improvement for Ai to detect compliance issues. It can also be expanded with new categories to detect.
Prompt should update ComplianceReader from a basic keyword filter to a sophisticated semantic tool. For a public institution like TUM, legal risks often hide in the subtext of job postings. It still fails to catch "Coded Compliance errors" and phrases that don't use forbidden words but still create an exclusionary effect.
By updating prompt to a stricter Contextual Compliance Detection and strictly enforcing  a new Public Sector Mandates (WissZeitVG) category, this change should ensure that job postings are not only legally complete but also align with the university's commitment to diversity and data privacy.
### Tasks

_No response_

### Acceptance Criteria

_No response_

ge94zec:
- added DSGVO and PUBLIC SECTOR EQUALITY
- made prompt stricter and more sensitive to implicit discrimination

#### General

#### Server

- [x] I added multiple server tests (JUnit) related to the features (with a high test coverage), while following the [server test guidelines](https://ls1intum.github.io/tum-apply/developer/server-guidelines/server-tests).

### Motivation and Context

Closes: https://github.com/ls1intum/tum-apply/issues/2344

### Description

New categories “DSGVO” and “Public Sector Equality” have been added. DSGVO was previously missing and partly overlapping with the transparency category. This change ensures that both aspects are now clearly distinguished and detected as separate categories. The sensitivity for implicit and contextual discrimination (e.g. phrases like “young team” or “dynamic team”) has been increased.
Based on research, universities offering PhD positions are part of the public service recruitment system and must therefore comply with public sector equality requirements.
This includes mandatory statements regarding disability inclusion and gender equality (§ 165 SGB IX). If such statements are missing, this may be considered a disadvantage to protected minority groups.

### Steps for Testing

Prerequisites:

1. Log in to Postman
2. Go to Postman
3. Fill in the URL with http://localhost:8080/api/ai/analyzeJobDescription?lang=de
4. Set content-type application/json
5. Verify with different texts in body if non compliant snippets are added to issues 

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

#### Server

| ComplianceCategory.java | 100.00% | 7 |

_Last updated: 2026-05-01 10:10:15 UTC_


src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,53 @@
...
+   If a text implies SUBTEXT and EXCLUSIONARY EFFECT, it is a violation
+----------------------
+3. CATEGORY CRITICAL_AGG:
   - Detect discrimination by protected characteristics under section 1 AGG.
   - Detect exclusion based on disability, age, origin, or gender.
   - Detect requests for unnecessary personal data or sharing with third parties without consent under Art. 5 DSGVO.
   - JOB TITLE CHECK: If title "{title}" is not gender-neutral with "(m/f/d)" or "(m/w/d)", flag it.
-  2.
+----------------------
+4. DSGVO MINIMIZATION:
```
Cathy0123456789: You always wrote "Category" before the other headers, I guess it's missing here then?

Cathy0123456789 (changes requested):
Could you please adapt the enum ComplianceCategory to reflect the new categories? I think the parsing of the response into the ComplianceIssue might not work properly otherwise

ge94zec:
> Could you please adapt the enum ComplianceCategory to reflect the new categories? I think the parsing of the response into the ComplianceIssue might not work properly otherwise



> Could you please adapt the enum ComplianceCategory to reflect the new categories? I think the parsing of the response into the ComplianceIssue might not work properly otherwise

This will be tackled in the following PR when I add the new categories in the sidebar :) 
This one is mainly for prompt

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,53 @@
...
+  - NOT triggered by: When explicit DATA is requested from the applicant.
+  - action: Suggest a message that clarifies whether applicant data is shared with external recipients (Art. 13 DSGVO)
+----------------------
+6. CATEGORY PUBLIC_SECTOR_EQUALITY: (SGB IX, BayGlG, WissZeitVG)
+  - Institutional Mandates: Public universities MUST include clauses for disability preference and female promotion.
+  - PhD Specific: For PhD/Academic positions, the text should imply that the position serves the purpose of "scientific qualification" or "promotion" (Wissenschaftliche Qualifizierung/Promotion).
+  - Semantic Existence Check: You must verify if the FOLLOWING THREE CONCEPTS are present in the text (not just keywords, but the intent):
+      1. Disability Notice: Statement that severely disabled applicants are given preference in the event of equal suitability (SGB IX).
+      2. Equality Notice: Statement that the university aims to increase the proportion of women and explicitly encourages women to apply (BayGlG).
+      3. Qualification Purpose: Mention that the contract is for obtaining a PhD/further qualification.
```
Cathy0123456789: The whole website is only for applying to PhD positions, do we really need to add this? @az108

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,53 @@
...
+  - Detect mentions of THIRD PARTIES/TOOLS (Workday, Headhunter, Partner Uni, consortia) without stating that data is shared with them.
+  - NOT triggered by: When explicit DATA is requested from the applicant.
+  - action: Suggest a message that clarifies whether applicant data is shared with external recipients (Art. 13 DSGVO)
+----------------------
+6. CATEGORY PUBLIC_SECTOR_EQUALITY: (SGB IX, BayGlG, WissZeitVG)
+  - Institutional Mandates: Public universities MUST include clauses for disability preference and female promotion.
+  - PhD Specific: For PhD/Academic positions, the text should imply that the position serves the purpose of "scientific qualification" or "promotion" (Wissenschaftliche Qualifizierung/Promotion).
+  - Semantic Existence Check: You must verify if the FOLLOWING THREE CONCEPTS are present in the text (not just keywords, but the intent):
+      1. Disability Notice: Statement that severely disabled applicants are given preference in the event of equal suitability (SGB IX).
+      2. Equality Notice: Statement that the university aims to increase the proportion of women and explicitly encourages women to apply (BayGlG).
```
Cathy0123456789: This is not what the professor needs to include in the job description. For that we have this statement automatically appended to every job:

Image

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,53 @@
...
-- Action: Suggest a message that clarifies whether applicant data is shared with external recipients (Art. 13 DSGVO)
-
+  - Detect for keyword that specifically ask for DATA (e.g., photos, ID copies, criminal records, marital status, religion)
+----------------------
+5. CATEGORY TRANSPARENCY:
+  - Detect mentions of THIRD PARTIES/TOOLS (Workday, Headhunter, Partner Uni, consortia) without stating that data is shared with them.
+  - NOT triggered by: When explicit DATA is requested from the applicant.
+  - action: Suggest a message that clarifies whether applicant data is shared with external recipients (Art. 13 DSGVO)
+----------------------
+6. CATEGORY PUBLIC_SECTOR_EQUALITY: (SGB IX, BayGlG, WissZeitVG)
```
Cathy0123456789: I think this whole sector becomes redundant then, if the previous things are removed

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,53 @@
...
-- category must be exactly CRITICAL_AGG or TRANSPARENCY.
-- action must be exactly REPLACE or ADD or REMOVE.
-- text: MAXIMUM 2-4 WORDS. Extract exact snippet from input text that uniquely identifies the violation (plain text, no HTML tags).
-- explanation must be a single SHORT sentence stating the legal violation directly. Sound like a formal legal notice.
-- If no issues exist, return exactly [].
+  - Return ONLY a valid JSON array. No markdown. No prose.
+  - One object per issue
+  - Object fields must be exactly:
+    id, text, category, article, explanation, action
+  - category must be exactly CRITICAL_AGG, TRANSPARENCY, DSGVO_MINIMIZATION und PUBLIC_SECTOR_EQUALITY
```
Cathy0123456789: I guess you meant to write "or", and not "und"?

Cathy0123456789 (changes requested):
Some comments still

Cathy0123456789:
> > Could you please adapt the enum ComplianceCategory to reflect the new categories? I think the parsing of the response into the ComplianceIssue might not work properly otherwise
> 
> > Could you please adapt the enum ComplianceCategory to reflect the new categories? I think the parsing of the response into the ComplianceIssue might not work properly otherwise
> 
> This will be tackled in the following PR when I add the new categories in the sidebar :) This one is mainly for prompt

However, when the AI returns one of the new categories, parsing the result will result in an exception. I tested this as well:
image

I don't think we want this even as intermediate result. This could be avoided by adding the categories to the enum, even if you don't show them yet on the client side or anything, but then at least we have no errors

src/main/resources/prompts/AnalyzeComplianceText.st
ge94zec: Yeah it is pretty much a big deal, after I found it on my research.

> Based on research, universities offering PhD positions are part of the public service recruitment system and must therefore comply with public sector equality requirements.
> This includes mandatory statements regarding disability inclusion and gender equality (§ 165 SGB IX). If such statements are missing, this may be considered a disadvantage to protected minority groups.

Regarding the Qualification Purpose, apparently every job description needs to make clear that the position's purpose serves scientific qualification. So it can't just be in a heading or as a separate field somewhere. it needs to be embedded in the actual job description text. A standalone label like "PhD Position" or a metadata tag isn't enough; the qualification purpose has to be part of the descriptive prose so the legal context is clearly conveyed to the reader. It's more of a rhetorical device.  I've often seen people only adding links referring to their webpages, but since it's an academic contract for a limited time, words like promotion or Wissenschaftliche Qualifizierung make the qualification context explicit.

src/main/resources/prompts/AnalyzeComplianceText.st
ge94zec: It has to be in the job description, this is very important. You are right, it is already included, but especially in underrepresented positions it could be better to have more personal or specific encouraging statements. AI could be a great tool to achieve that and make inclusion even more present.

src/main/resources/prompts/AnalyzeComplianceText.st
Cathy0123456789: My comment was only on the last line (point 3)

src/main/resources/prompts/AnalyzeComplianceText.st
Cathy0123456789: But it is already since we append this to every job, or am I misunderstanding something?

Cathy0123456789: This is also what Stephan mentioned at the Lunchtalk to the other professors, that they do not need to write this into every job description, since we take care of this for them. And that's why also Aniruddh adjusted the text according to what TUM wants so we are complying

src/main/resources/prompts/AnalyzeComplianceText.st
ge94zec: Ah I think I get now what makes it feel misleading or redundant. So to summarize simply, the other categories (Inclusion, AGG, DSGVO) analyze the language and wording of the text, whether no discriminatory phrasing is used, etc. But PUBLIC_SECTOR_EQUALITY is different, it's an existence check for two specific mandatory sentences that German public universities are legally required to include in every job posting. These are covered by the disability statement under § 165 SGB IX and the equality statement under BayGlG, like the prompt says. So every indivudell "Stellenbeschreibung" needs to include these two sentences, unfortunately, the export pdf does not include these statements. To legally secure us, it should be present in the job description text itself. It also feels more sincere when added in job description itself rather than as a generic statement beneath. This is very important and standard in universities and PhD recruitment processes in Germany. Hope this explanation resolves the misunderstanding :))

src/main/resources/prompts/AnalyzeComplianceText.st
ge94zec: Thanks I will change it ^^

src/main/resources/prompts/AnalyzeComplianceText.st
Cathy0123456789: Alright makes sense 👍🏼

src/main/resources/prompts/AnalyzeComplianceText.st
Cathy0123456789: Personally, I think asking professors to include this again is going to be too much, but that maybe @az108 can decide. My feeling is if AI points out too much or these more "optional" things, it might get overloaded for professors, it would be rather more important for pointing out mistakes and things that might be a compliance risk. If it would be worded excludingly towards women for example, then it should be flagged definitely, similarly to what the gender decoder does, but explicitly asking professors to word it to include women and disabled people might be more deterrent (like we don't use gender-exclusive but rather non-gender-inclusive or we don't use "männlich and weiblich kodiert" for that reason). Also, Stephan mentioned reaching 100% should not be too difficult. But this is just a personal feeling, could be wrong as well

src/main/resources/prompts/AnalyzeComplianceText.st
Cathy0123456789: In this case I think it makes more sense to include these parts in the export rather than forcing this for every job description. I think legally that should be fine as well, since this was already an issue resolved from TUM side some months ago, so the status quo seems to be fine

Cathy0123456789: This is the current job portal from TUM itself: https://portal.mytum.de/jobs/wissenschaftler
They do the same, they add it just underneath the description the way we do it currently

src/main/resources/prompts/AnalyzeComplianceText.st
az108: i agree with cathy, and please remove the disability and equlity notice from the prompt as this is duplication and just decreases ai output (the more you put in the prompt the worse the output)

src/main/resources/prompts/AnalyzeComplianceText.st
ge94zec: > In this case I think it makes more sense to include these parts in the export rather than forcing this for every job description. I think legally that should be fine as well, since this was already an issue resolved from TUM side some months ago, so the status quo seems to be fine



> In this case I think it makes more sense to include these parts in the export rather than forcing this for every job description. I think legally that should be fine as well, since this was already an issue resolved from TUM side some months ago, so the status quo seems to be fine

Okay, i will then add a new issue that includes these parts in the export.

src/main/java/de/tum/cit/aet/ai/constants/ComplianceCategory.java
```diff
@@ -3,6 +3,6 @@
 public enum ComplianceCategory {
     CRITICAL_AGG,
     TRANSPARENCY,
-    GENDER_INCLUSIVE,
-    GENDER_EXCLUSIVE,
+    DSGVO_MIN,
+    PUBLIC_SELECTOR,
```
Cathy0123456789: I guess you meant to write "PUBLIC_SECTOR" rather than "SELECTOR"?

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,49 @@
-You are the TUMApply ComplianceReader, an AI legal and compliance assistant for academic job postings at a prestigious university.
+You are the TUMApply ComplianceReader, an advanced AI legal and compliance expert for academic job postings at a prestigious university.
+Your core capability is "Contextual Compliance Detection" based on German and EU law.
+You analyze the semantic intent and the exclusionary effect of sentences, not just keywords.
 
 TASK:
-Analyze the provided job posting text for legal risks and missing information based on German and EU law.
+Deeply analyze the job posting for legal risks (AGG, DSGVO, SGB IX, BayGlG) based on German and EU law.
```
Cathy0123456789: I guess SGB IX and ByGlG can be removed here as well?

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,49 @@
...
+  - Detect for keyword that specifically ask for DATA (e.g., photos, ID copies, criminal records, marital status, religion)
+----------------------
+5. CATEGORY TRANSPARENCY:
+  - Detect mentions of THIRD PARTIES/TOOLS (Workday, Headhunter, Partner Uni, consortia) without stating that data is shared with them.
+  - NOT triggered by: When explicit DATA is requested from the applicant.
+  - action: Suggest a message that clarifies whether applicant data is shared with external recipients (Art. 13 DSGVO)
+----------------------
+6. CATEGORY PUBLIC_SECTOR_EQUALITY: (WissZeitVG)
+  - PhD Specific: For PhD/Academic positions, the text should imply that the position serves the purpose of "scientific qualification" or "promotion" (Wissenschaftliche Qualifizierung/Promotion).
+  - Semantic Existence Check: You must verify if  a qualification purpose is present in the text (not just keywords, but the intent)
```
Cathy0123456789: Do this sentence and the previous not kind of mean the same?

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,41 +1,49 @@
...
-- category must be exactly CRITICAL_AGG or TRANSPARENCY.
-- action must be exactly REPLACE or ADD or REMOVE.
-- text: MAXIMUM 2-4 WORDS. Extract exact snippet from input text that uniquely identifies the violation (plain text, no HTML tags).
-- explanation must be a single SHORT sentence stating the legal violation directly. Sound like a formal legal notice.
-- If no issues exist, return exactly [].
+  - Return ONLY a valid JSON array. No markdown. No prose.
+  - One object per issue
+  - Object fields must be exactly:
+    id, text, category, article, explanation, action
+  - category must be exactly CRITICAL_AGG, TRANSPARENCY, DSGVO_MINIMIZATION or PUBLIC_SECTOR_EQUALITY
```
Cathy0123456789: Can you please adapt either the prompt or the enum so both have the exact wording otherwise the parsing will fail

Cathy0123456789 (changes requested):
Thank you, small comments still

Cathy0123456789 (approved):
Thank you, code LGTM, not tested though since TS is down

az108 (approved):

PR #2404, opened 2026-05-02
`Bugfix`: Resolve 401 Unauthorized on Refresh

az108:
#### General

#### Server

#### Client

### Motivation and Context

Closes #2403.

When a user opens the app while not logged in, the client fires a startup probe to `POST /api/auth/refresh` (`auth-facade.service.ts:66`) to decide whether to restore a server session or fall back to Keycloak SSO. With no session cookies present, the server returned `401 Unauthorized`, which surfaces as a red error in browser dev tools and HTTP-level monitoring even though the situation is fully expected.

### Description

- `AuthSessionInfoDTO` gains an `authenticated` flag (defaults to `true` for existing call sites).
- `AuthenticationResource#refresh` now always returns `200`:
  - **No auth cookies present** → `AuthSessionInfoDTO.unauthenticated()` (the anonymous probe case).
  - **Cookies present but Keycloak rejects refresh** → cookies are cleared and the same unauthenticated payload is returned.
  - **Refresh succeeds** → unchanged: cookies are rotated and lifetimes are returned with `authenticated=true` (default).
- `ServerAuthenticationService#refreshTokens` treats `authenticated=false` as "no server session" without `console.warn` or toast (initialization probe stays silent).
- 401 is still returned by every other authenticated endpoint — only the probe contract changed.

### Steps for Testing

Prerequisites:

1. Have the server and client running locally without an existing session (clear cookies / use a private window).

Test cases:

1. Open the app while logged out → DevTools should show `POST /api/auth/refresh` returning **200** with `{"authenticated": false, ...}`. No red 401 error.
2. Log in via email/password → cookies are set; subsequent refresh should still rotate the session normally and `authenticated` is omitted/`true` in the response.
3. Manually expire/invalidate the refresh cookie (e.g. tamper with its value) and reload → the response is still 200 with `authenticated=false`, and the stale cookies are cleared as a side effect.
4. Confirm that a non-public endpoint (e.g. `/api/users/me`) still returns 401 when called without a valid session — only the refresh probe contract is loosened.

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Anonymous probe returns 200 with `authenticated=false`
- [ ] Authenticated refresh still rotates cookies and returns lifetimes
- [ ] Stale cookie path clears cookies and returns `authenticated=false`
- [ ] Other protected endpoints still return 401 when unauthenticated


Cathy0123456789 (approved):
Tested locally, I see no refresh errors anymore

PR #2406, opened 2026-05-02
`Bugfix`: Fix frozen Login button on slow networks

### LINKED ISSUE #2405: Login button does nothing on slow networks or after logout
### Summary

Sometimes when the page is loaded — or right after logging out — clicking the Login button does nothing and the UI feels frozen. The button does not show an error and gives no visual feedback.

### How to reproduce

1. Open TUMApply on a slow / unstable network (or with the Keycloak server responding slowly).
2. As soon as the page is interactive, click "Login".
   _or_
   Log out, then immediately click "Login" again.
3. Observe: the button click is silently rejected. Nothing opens, nothing errors.

### Why it happens

`AuthFacadeService.initAuth()` runs at app startup to silently restore an existing session (server refresh + Keycloak `check-sso`). It was wrapped in `runAuthAction`, which sets `authOrchestrator.isBusy = true` for the whole duration. Any user-driven auth action (including clicking Login) checks `isBusy()` first and throws "AuthOrchestrator is busy" if it is true.

On a slow network, the silent SSO iframe or the refresh call can take several seconds — and during that window every Login click is silently dropped.

### Expected behaviour

- Clicking Login is never blocked by the silent background SSO/refresh bootstrap.
- A slow Keycloak silent-SSO check must not freeze the Login UI.

### Fix

PR #2406 decouples `initAuth` from the user-facing busy lane. It runs as a real background task with its own error handling and no longer toggles `isBusy`. Guards prevent it from clobbering a session that a user-driven login may have established in the meantime.

az108:
#### General

#### Client

### Motivation and Context

Closes #2405.

On slow or unstable networks — and sometimes right after logging out — clicking the Login button does nothing. No toast, no error, nothing opens. Users perceive the page as frozen.

Root cause: `AuthFacadeService.initAuth()` was wrapped in `runAuthAction`, which means the silent SSO / refresh bootstrap that runs at app startup sets `authOrchestrator.isBusy = true` for the entire duration of its work (server refresh + Keycloak `check-sso` iframe). User-initiated auth actions guard on `isBusy()` and silently throw "AuthOrchestrator is busy". On a slow Keycloak / slow network, that window is large enough for users to click Login and get nothing.

### Description

- `initAuth` no longer goes through `runAuthAction`. It is a silent background bootstrap, not a user-initiated action, so it must not toggle the user-facing busy flag. The `isBusy` lane now belongs only to explicit user actions (Login, OTP, Logout, Register).
- Added `authMethod !== 'none'` guards after every `await` inside `initAuth`, so a slow init that finishes after a user-driven login cannot clobber the session that the user just established.
- Errors during init no longer rethrow. They surface a toast and log a warning, but resolve to `false`. `initAuth` is awaited by Angular's `provideAppInitializer`, so a rejected promise there would block app bootstrap. This mirrors the existing pattern in `KeycloakAuthenticationService.init()`.
- Added two unit tests:
  - `initAuth` does not toggle `authOrchestrator.isBusy`.
  - `loginWithEmail` succeeds while a slow Keycloak `init()` is still in flight, and the late-resolving init does not overwrite `authMethod`.

### Steps for Testing

Prerequisites:

1. Run the client locally against a working Keycloak.
2. In Chrome DevTools, open the Network tab and throttle the connection (or use a custom profile that adds 3–5 s latency to the Keycloak host).

Test 1 — slow load:

1. Reload the page with the throttled network active.
2. As soon as the page is interactive, click **Login**. The login dialog should open immediately, even while the Keycloak silent-SSO check is still in flight in the background.

Test 2 — logout into login:

1. Log in with email/password.
2. Log out.
3. Immediately click **Login** again. The login dialog should open without delay, no silent rejection.

Test 3 — race protection:

1. Throttle Keycloak as in Test 1.
2. Reload, then immediately log in with email/password.
3. After the slow Keycloak silent-SSO check finally resolves, the user must remain authenticated as the server-session user — the late init must not flip the active auth method to Keycloak.

Automated:

- `npx vitest run src/test/webapp/app/core/auth/auth-facade.service.spec.ts` — all 16 tests pass, including the two new ones covering the scenarios above.

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1 — slow load, Login button responsive
- [ ] Test 2 — logout then Login, no freeze
- [ ] Test 3 — server session is not overwritten by late silent SSO

### DIFF
--- a/src/test/webapp/app/core/auth/auth-facade.service.spec.ts
+++ b/src/test/webapp/app/core/auth/auth-facade.service.spec.ts
@@ -79,6 +79,46 @@ describe('AuthFacadeService', () => {
       const result = await facade.initAuth();
       expect(result).toBe(false);
     });
+
+    it('does not toggle orchestrator isBusy (background bootstrap)', async () => {
+      const { facade, server, keycloak, orchestrator } = setup();
+      let observedDuringRefresh = false;
+      server.refreshTokens.mockImplementation(async () => {
+        observedDuringRefresh = orchestrator.isBusy();
+        return false;
+      });
+      keycloak.init.mockResolvedValue(false);
+
+      await facade.initAuth();
+
+      expect(observedDuringRefresh).toBe(false);
+      expect(orchestrator.isBusy()).toBe(false);
+    });
+
+    it('does not block user-initiated login while in flight', async () => {
+      const { facade, server, keycloak, account } = setup();
+
+      let resolveKeycloakInit: (v: boolean) => void = () => {};
+      const keycloakInitPromise = new Promise<boolean>(resolve => {
+        resolveKeycloakInit = resolve;
+      });
+      server.refreshTokens.mockResolvedValue(false);
+      keycloak.init.mockReturnValue(keycloakInitPromise);
+      account.loadUser.mockResolvedValue(undefined);
+
+      // Kick off init but don't await — simulates a slow Keycloak silent SSO check.
+      const initPromise = facade.initAuth();
+
+      // While init is still in flight, the user clicks Login. It must succeed.
+      server.login.mockResolvedValue(undefined);
+      await expect(facade.loginWithEmail('a@b.com', 'pw')).resolves.toBe(true);
+      expect(server.login).toHaveBeenCalledTimes(1);
+
+      // Finish the slow init; it must not overwrite the server-established session.
+      resolveKeycloakInit(true);
+      await initPromise;
+      expect((facade as any).authMethod).toBe('server');
+    });
   });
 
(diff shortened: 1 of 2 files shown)


src/test/webapp/app/core/auth/auth-facade.service.spec.ts
Cathy0123456789, line 1: Can we please start test case names with "should" 👀

Cathy0123456789 (changes requested):

CelineLahnor (approved):
This looks good from my side thank you

PR #2411, opened 2026-05-02
`Bugfix`: Fix PDF export footer timestamp and link border

az108:
#### General

#### Server

> Note on server tests: this is a presentational PDF fix (timestamp formatting + iText link annotation border). The existing service has no unit tests around the metadata footer because the surrounding helpers also do not (see `formatToBerlinDate` etc.), and unit-testing iText annotation flags is fragile. Verified via `./gradlew compileJava` and the manual steps below.

### Motivation and Context

Closes #2410.

QA reported two visual issues with the metadata footer of exported PDFs (\"This document was created on … by … using TumApply.\"):

1. The timestamp is one hour behind the user's local time — on the production JVM (UTC) the footer renders UTC, not Munich time, with no zone indicator.
2. The \"TumApply\" hyperlink is wrapped in a thin blue outline rectangle on top of the underlined text.

### Description

- **Timestamp:** `PDFExportService.buildMetadataText` used `LocalDateTime.now()` without a zone. The same service already reads instants in `ZoneId.of(\"Europe/Berlin\")` for every other date helper (`formatToBerlinDate`, `formatToBerlinTime`, `formatToBerlinDateTime`). Aligned the footer to use the same zone so the timestamp is always Munich local time.
- **Link border:** iText's `Link` element creates a `PdfLinkAnnotation` whose default border is a 1pt solid line. That renders as a blue outline rectangle around \"TumApply\" on top of our underline. Calling `tumapplyLink.getLinkAnnotation().setBorder(new PdfArray(new float[] { 0, 0, 0 }))` removes only the annotation border; the styling on the `Link` element (colour, underline, font) is preserved.

### Steps for Testing

Prerequisites:

1. Start the server locally.
2. Log in.

Test 1 — applicant PDF export (timestamp + link):

1. As an applicant, export an application as PDF.
2. Open the PDF and scroll to the metadata footer.
3. The timestamp matches your Munich wall clock (within a minute).
4. The word \"TumApply\" is shown only as underlined coloured text — no rectangle around it.

Test 2 — professor PDF export:

1. As a professor, export a job as PDF.
2. Same checks as in Test 1.

Test 3 — production-like UTC JVM:

1. Run the server with `-Duser.timezone=UTC` (or in a Docker image whose container TZ is UTC).
2. Export a PDF and confirm the footer still shows Munich local time, not UTC.

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [ ] Test 1 — applicant export: timestamp is local, no link box
- [ ] Test 2 — professor export: same
- [ ] Test 3 — UTC JVM: timestamp still shows Munich time

### DIFF
--- a/src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
+++ b/src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
@@ -480,7 +480,11 @@ public class PDFExportService {
     private String buildMetadataText(Map<String, String> labels) {
         StringBuilder metadata = new StringBuilder();
 
-        String currentDateTime = LocalDateTime.now().format(DATETIME_FORMATTER);
+        // Render the footer timestamp in Munich time. Production JVMs run on UTC, so a plain
+        // LocalDateTime.now() footer is one hour behind during CET (or two during CEST), which
+        // is misleading for the user reading the PDF. Matches the pattern used by the other
+        // date helpers in this service.
+        String currentDateTime = LocalDateTime.now(ZoneId.of("Europe/Berlin")).format(DATETIME_FORMATTER);
         metadata.append(labels.get("thisDocumentWasGeneratedOn"));
         metadata.append(currentDateTime);
 
--- a/src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
+++ b/src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
@@ -47,8 +47,9 @@ public class PDFExportService {
     /**
      * Exports application details to PDF
      *
-     * @param app    the ApplicationDetailDTO containing application data
-     * @param labels translation labels for PDF content
+     * @param app      the ApplicationDetailDTO containing application data
+     * @param labels   translation labels for PDF content
+     * @param timezone IANA zone id used to render the footer "generated on" timestamp; falls back to Europe/Berlin when null/blank/invalid
      * @return the PDF file as Resource
      */
     public Resource exportApplicationToPDF(ApplicationDetailDTO app, Map<String, String> labels, String timezone) {
@@ -159,8 +160,9 @@ public class PDFExportService {
     /**
      * Exports job details to PDF
      *
-     * @param jobId  the job ID
-     * @param labels translation labels for PDF content
+     * @param jobId    the job ID
+     * @param labels   translation labels for PDF content
+     * @param timezone IANA zone id used to render the footer "generated on" timestamp; falls back to Europe/Berlin when null/blank/invalid
      * @return the PDF file as Resource
      */
     public Resource exportJobToPDF(UUID jobId, Map<String, String> labels, String timezone) {
@@ -228,6 +230,7 @@ public class PDFExportService {
      *
      * @param jobFormDTO the job form data
      * @param labels     translation labels for PDF content
+     * @param timezone   IANA zone id used to render the footer "generated on" timestamp; falls back to Europe/Berlin when null/blank/invalid
      * @return the PDF file as Resource
      */
     public Resource exportJobPreviewToPDF(JobFormDTO jobFormDTO, Map<String, String> labels, String timezone) {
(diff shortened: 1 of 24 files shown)


src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
```diff
@@ -480,7 +480,7 @@ public String generateApplicationFilename(String jobTitle, String applicationLab
     private String buildMetadataText(Map<String, String> labels) {
         StringBuilder metadata = new StringBuilder();
 
-        String currentDateTime = LocalDateTime.now().format(DATETIME_FORMATTER);
+        String currentDateTime = LocalDateTime.now(ZoneId.of("Europe/Berlin")).format(DATETIME_FORMATTER);
```
CelineLahnor: Maybe we could use the user’s local ZoneId, so the displayed time always matches their current local time, no matter where they are?

CelineLahnor (changes requested):
Thank you I left a small suggestion :)

src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
CelineLahnor, line 303: exportInterviewToPDF(...) has no timezone parameter, so interview PDFs cannot become viewer-local through the same path as the other exports.

src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
```diff
@@ -344,7 +349,7 @@ public Resource exportInterviewToPDF(Interviewee interviewee, Application app, J
         builder.startInfoSection(labels.get("interviewNotes")).addSectionContent(getValue(interviewee.getAssessmentNotes()));
 
         // Footer
-        builder.setMetadata(buildMetadataText(labels));
+        builder.setMetadata(buildMetadataText(labels, null));
```
CelineLahnor: What is the intention to ignore the viewers timezone here?

src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
CelineLahnor, line 371: Here is still hardcode ZoneId.of("Europe/Berlin")

CelineLahnor: Here as well

CelineLahnor: Same here

src/main/java/de/tum/cit/aet/core/service/PDFExportService.java
```diff
@@ -472,15 +477,35 @@ public String generateApplicationFilename(String jobTitle, String applicationLab
...
+     * @param timezone IANA zone id used to render the "generated on" timestamp;
+     *                 falls back to Europe/Berlin when null/blank/invalid
      * @return formatted metadata string
      */
-    private String buildMetadataText(Map<String, String> labels) {
+    private String buildMetadataText(Map<String, String> labels, String timezone) {
         StringBuilder metadata = new StringBuilder();
 
-        String currentDateTime = LocalDateTime.now().format(DATETIME_FORMATTER);
+        String currentDateTime = LocalDateTime.now(resolveZone(timezone)).format(DATETIME_FORMATTER);
```
CelineLahnor: Can you please use this logic for the interview helpers as well?

CelineLahnor (changes requested):
We now use viewer-local timezones, which is the right direction, but this needs to be applied consistently across the whole PDF. Otherwise, users downloading from another timezone will get mixed-zone timestamps in the same document.

Cathy0123456789:
> We now use viewer-local timezones, which is the right direction, but this needs to be applied consistently across the whole PDF. Otherwise, users downloading from another timezone will get mixed-zone timestamps in the same document.

I disagree, I think no user is going to actually look at it. Also, we internally store it in UTC and it is a standard. The JSON export would then also just use UTC and everything is consistent. I would write it in the PDF that the date/time is UTC and that should be fine then. Also, that saves us a lot of unnecessary code.

CelineLahnor (approved):
This is good for me

Cathy0123456789 (approved):
Tested locally, PDF looks good, thank you. JSON doesn't show UTC time for me, rather my local time, but can be checked follow up

PR #2450, opened 2026-05-03
`General`: Change the subject area selector visibility to always be visible

### LINKED ISSUE #2330: Change the subject area selector visibility to always be visible
### Description

Currently the subject area selector visible only when the “Receive Notifications” toggle is on. We want to keep the subject area selector visible even when notifications are disabled, so users can review or prepare their preferences in advance.

### Tasks

- Make the subject area selector visible even when the “Receive Notifications” toggle is off
- Adjust the info text accordingly

### Acceptance Criteria

_No response_

CelineLahnor:
#### General

#### Client

### Motivation and Context

Closes #2330 

### Description

This PR changes the subject area selector to always be visible even when notifications are disabled, so users can review or prepare their preferences in advance. 

1. Log in as applicant
2. Go to settings -> notifications
3. Enable and Disable the notifications settings and verify the selector always stays visible

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

#### Client

| notification-settings.component.ts | 100.00% | 180 | 53 | 29.4 |

_Last updated: 2026-05-03 15:58:29 UTC_


src/main/webapp/app/shared/settings/notifications/notification-settings.component.html
```diff
@@ -38,21 +38,16 @@ <h2 class="font-medium text-text-primary text-lg" jhiTranslate="settings.notific
...
-                  <jhi-subject-area-subscription-selector
-                    [saving]="subjectAreaSubscriptions.saving()"
-                    [filterOptions]="subjectAreaSubscriptions.filterOptions"
-                    [selectedValues]="subjectAreaSubscriptions.selectedFilterValues()"
-                    [selectedOptions]="subjectAreaSubscriptions.selectedOptions()"
-                    (filterChange)="subjectAreaSubscriptions.onFilterChange($event)"
-                    (removeSubjectArea)="subjectAreaSubscriptions.remove($event)"
-                  />
-                </div>
+            <div class="mt-4 flex flex-col gap-3">
```
Cathy0123456789: flex please at the start

src/main/webapp/app/shared/settings/notifications/notification-settings.component.html
```diff
@@ -38,21 +38,16 @@ <h2 class="font-medium text-text-primary text-lg" jhiTranslate="settings.notific
...
+              <div [class.pointer-events-none]="subjectAreaSubscriptions.saving()">
```
Cathy0123456789: Is this needed?

Cathy0123456789 (changes requested):
Thank you, small comments

src/main/webapp/app/shared/settings/notifications/notification-settings.component.html
CelineLahnor: Without pointer-events-none, the UI would remain clickable while the save is in progress.

Cathy0123456789: Which should be fine or not? I tested it and it doesn't seem to break anything. The save gets done just as normal.
Also, isn't it strange, if the user tries to click something and the dropdown closes instead? There is no indicator that saving is still ongoing, so it creates a strange effect

CelineLahnor: It’s mainly to avoid concurrent updates. If the user clicks again while the first save is still in flight, we can end up with overlapping requests based on stale state, and the final result may depend on request timing. So it’s not strictly needed, but it is a defensive safeguard against race conditions. @az108 should I keep it or not?

az108: I agree with Cathy, let's drop it. The store already handles concurrency: updateSelection is optimistic with rollback + load() resync on failure, so the final state always converges to the server. Plus saving isn't ref-counted, so the guard only blocks during the very first request anyway, and as Cathy noticed it makes the dropdown collapse on click instead of registering, worse UX than no guard. If we ever need real serialization later, that belongs in the store, not as a CSS gate.

az108 (approved):

Cathy0123456789 (approved):
Thank you

PR #2470, opened 2026-05-05
`Development`: Reuse compliance analysis across languages via snippet mapping 

### LINKED ISSUE #2346: Enhance performance for compliance
### Description

Make performance faster and add streaming for highlights. 

### Tasks

_No response_

### Acceptance Criteria

_No response_

ge94zec:
- added endpoint POST that maps source issues to a target language
- added prompt that returns one mapped text snippet per line in source complianceIssues order, other fields carry over
- wired compliance mapping through client → server API
- separated compliance analysis from score calculation
- extracted save flow in JobService into Consumer
- improved performance by eliminating redundant analysis steps in job creation

#### General

#### Server

#### Client

### Motivation and Context

Closes: https://github.com/ls1intum/tum-apply/issues/2346

### Description

This change improves the bilingual AI compliance workflow for job descriptions by replacing the second  target-language compliance analysis with a snippet mapping. The editor now analyzes the source language once in  `analyzeJobDescription`, streams and stores the translated text in `translateTextStream`, and then maps the detected compliance snippets onto the translated version automatically `mapComplianceIssues`. This reduces duplicate AI work, keeps translated highlights in sync with sourceIssues. 

The duplicated compliance issue persistence logic in JobService was extracted into a shared helper. Since the Score is currently a combined value derived from both gender and legal analysis, this ensures the score remains consistent and is only updated when a full, valid recalculation is possible.

LLM compliance calls per autosave cycle is now reduced from 2 to 1. The translation with second analysis is reduced from 11s to approximately 2.30s.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply as Prof 
2. Nav to Create position 
3. Verify that the complianceAnalysis of target stays within a 2-second threshold

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [ ] Test 1

### Screenshots

Bildschirmfoto 2026-05-09 um 19 08 15

### Test Coverage

**Warning:** Some tests failed in the PR check. Coverage may be incomplete. Please check the [workflow logs](https://github.com/ls1intum/tum-apply/actions/runs/27284399749).

#### Client

| job-creation-form.component.ts | 67.31% | 1235 | 72 | 5.8 |

#### Server

| MapComplianceIssuesRequestDTO.java | not found (added) | 15 |
| AiService.java | not found (modified) | 335 |
| AiResource.java | not found (modified) | 99 |
| JobService.java | not found (modified) | 343 |

_Last updated: 2026-06-10 14:50:48 UTC_


src/main/java/de/tum/cit/aet/ai/dto/MapComplianceIssuesRequestDTO.java
```diff
@@ -0,0 +1,12 @@
+package de.tum.cit.aet.ai.dto;
+
+import de.tum.cit.aet.ai.domain.ComplianceIssue;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+
+public record MapComplianceIssuesRequestDTO(
```
Cathy0123456789, line 11: Can you please add a Json Include Annotation

src/main/java/de/tum/cit/aet/ai/web/AiResource.java
```diff
@@ -79,6 +82,30 @@ public ResponseEntity<Flux<String>> translateJobDescriptionStream(
...
+    @PostMapping(value = "map-compliance-issues", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<List<ComplianceIssue>> mapComplianceIssues(
+        @RequestParam("toLang") String toLang,
+        @RequestParam("jobId") UUID jobId,
+        @Valid @RequestBody MapComplianceIssuesRequestDTO request
+    ) {
+        if (!aiFeatureToggleService.isAiAvailable()) {
+            return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).build();
+        }
+        log.info("POST /api/ai/map-compliance-issues - Request received (toLang={})", toLang);
```
Cathy0123456789: Can we be a bit more specific than "Request received"

src/main/java/de/tum/cit/aet/ai/web/AiResource.java
```diff
@@ -79,6 +82,30 @@ public ResponseEntity<Flux<String>> translateJobDescriptionStream(
...
+        return ResponseEntity.ok(
+            aiService.mapComplianceIssues(jobId, request.complianceIssues(), request.text(), request.translatedText(), toLang)
```
Cathy0123456789: Instead of giving them as single parameters, would it not make sense to pass the DTO just?

src/main/resources/prompts/SnippetMapping.st
Cathy0123456789, line 1: Maybe adding an example is helpful?

Cathy0123456789 (changes requested):
Thank you, reviewed code

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,51 +1,59 @@
...
+
+OUTPUT
+  Return only a minified one-line JSON array; no markdown or prose. Use exactly these fields:
+  `text`, `category`, `suggestion`.
+  - `category`: CRITICAL_AGG, DSGVO_MINIMIZATION, TRANSPARENCY, or PUBLIC_SECTOR.
+  - `text`: exact shortest non-overlapping substring from the supplied title/description.
+  - `suggestion`: exact fix in {descriptionLanguage}; REPLACE = safe alternative, ADD = sentence to append, REMOVE = "".
+  - If compliant, return exactly [].
+
+  {format}
```
az108: `{format}` has no matching `.param(...)`. `AiService.analyzeJobDescription` passes only `descriptionLanguage`, `userLang`, `jobDescription` and `title`, and this PR does not touch that call site.

Spring AI's default renderer is `StTemplateRenderer` (`DefaultChatClient.java:88` — `DEFAULT_TEMPLATE_RENDERER = StTemplateRenderer.builder().build()`) with `DEFAULT_VALIDATION_MODE = ValidationMode.THROW` (`StTemplateRenderer.java:67`), and `SpringAIConfiguration.java:44` is a plain `ChatClient.builder(chatModel).build()` — no custom renderer, no relaxed validation. `.entity()` does **not** create a `format` template variable: `doSingleWithBeanOutputConverter` only puts the format into the request *context* under `ChatClientAttributes.OUTPUT_FORMAT`, and the user text was already rendered back at `.call()`. `format` is not an ST built-in either, so it is not skipped.

Reproduced against the real 2.0.0-M4 jars with both prompt versions and the four actual params:

```
PR   -> IllegalStateException: Not all variables were replaced in the template.
        Missing variable names are: [format].
MAIN -> RENDER OK, length=2895
```

The exception is caught by `analyzeJobDescription`, calls `aiFeatureToggleService.recordFailure()` (which also trips the AI circuit breaker) and surfaces as `InternalServerException("Compliance analysis parsing failed")`. It is not data-dependent — every invocation fails while the AI toggle is on. CI does not catch it: `AiResourceTest` mocks `AiService`, and nothing in the suite renders a real prompt.

Fix: drop this line. Spring AI appends the format instruction on its own, which is why no other prompt in the repo has `{format}`.

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,51 +1,59 @@
...
+    For PhD/academic positions, flag a missing statement that the role serves a doctorate or scientific qualification.
+    Do not flag if that purpose is already stated. Use the exact "Description" or "Beschreibung" heading as `text` when present;
+    otherwise use the exact title. Append one complete qualification-purpose sentence.
+
+INPUT
+  {title}
+  {jobDescription}
+
+OUTPUT
+  Return only a minified one-line JSON array; no markdown or prose. Use exactly these fields:
```
az108: The output contract no longer matches the type this is deserialized into. `analyzeJobDescription` still parses `List<ComplianceIssue>`, and `ComplianceIssue` has `id, category, text, article, explanation, action, language` — there is no `suggestion` field anywhere in `src/main/java` or the generated client. So `suggestion` is silently discarded by the converter (`FAIL_ON_UNKNOWN_PROPERTIES` is off), while `article`, `explanation` and `action` are no longer requested at all. `main` asked for them explicitly: *"Object fields must be exactly: id, text, category, article, explanation, action"*.

This is not cosmetic. `ai-compliance-popover.component.html:7-8` renders nothing but `{{ issue()?.article }}` and `{{ issue()?.explanation }}`, and `job-creation-form.component.ts:338` uses `issue.explanation` for `titleComplianceError`. Both end up empty or filled with whatever the model happens to invent from the JSON schema.

Either keep the previous field list, or add `suggestion` to `ComplianceIssue` and migrate the UI in the same PR.

src/main/resources/prompts/AnalyzeComplianceText.st
```diff
@@ -1,51 +1,59 @@
...
-  - If no issues exist, return exactly [].
+You are DocApply ComplianceReader for academic job postings. Reason briefly but thoroughly and return every finding in the first response.
+
+STRICT POLICY OVERRIDE — HIGHEST PRIORITY
+- The rules below define mandatory violations and override general legal knowledge or contextual interpretation.
+- Apply every stated trigger exactly. Only explicitly stated exceptions are valid.
+- Match meaning case-insensitively; examples are non-exhaustive.
+- Check every occurrence against all four categories and return every match. Do not narrate the analysis.
+
+LANGUAGES
```
az108: `{userLang}` is gone from the template — `main` had `The explanation must be written in: {userLang}.` here — but the parameter is still threaded through the entire chain: the client sends it (`job-creation-form.component.ts:1874`), `AiResource.analyzeJobDescription` takes `@RequestParam(defaultValue = "en") String userLanguage`, and `AiService` still calls `.param("userLang", userLang)`. The JavaDoc still states it "controls the language of explanation texts in the returned issues".

`StTemplateRenderer.validate` only reports *missing* variables, never surplus ones, so this fails silently — the user-language feature is simply gone while the API signature and docs still promise it. Either wire the variable back into the prompt or remove the parameter from client, resource, service and JavaDoc.

src/main/java/de/tum/cit/aet/ai/dto/MapComplianceIssuesRequestDTO.java
```diff
@@ -0,0 +1,16 @@
...
+import de.tum.cit.aet.ai.domain.ComplianceIssue;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import java.util.UUID;
+
+@JsonInclude
+public record MapComplianceIssuesRequestDTO(
+    String toLang,
+    UUID jobId,
```
az108: `jobId` is the only field here without validation, while `translatedText` and `complianceIssues` carry `@NotBlank` / `@NotNull`. `applyJobChangeForAnalysis` opens with `if (jobId == null) { return; }` — no log, no exception. A request without it runs the expensive LLM mapping, returns 200 with the mapped issues, and persists nothing. The empty-source-issues early return is affected too. The generated OpenAPI model has `readonly jobId?: string`, so a spec-compliant caller is allowed to omit it.

`@NotNull UUID jobId` is what the server guidelines document — the input-DTO example in `server-development.mdx` (*Request Validation*) is literally `public record CreateApplicationDTO(@NotBlank String motivation, @NotNull UUID jobId) {}` — and what `AssignSlotRequestDTO`, `BookSlotRequestDTO` and `UpdateApplicationDTO` already do.

az108 (changes requested):
Reviewed on top of Cathy's existing review — none of the points below overlap with hers.

The core idea is right: snippet mapping instead of a second full-text analysis is exactly the lever worth pulling here, and `applyJobChangeForAnalysis` quietly adds the *first* ownership check on this path — `updateAiAnalysis` had none on `main`. Raw `List` → `List<ComplianceIssue>` is a good catch too.

The prompt rewrite that ships with it is broken, though.

### Blocking

**1. `{format}` breaks every compliance analysis** (`AnalyzeComplianceText.st:59`) — unresolved template variable, `StTemplateRenderer` defaults to `ValidationMode.THROW`. Reproduced against the real Spring AI 2.0.0-M4 jars; details inline.

**2. Output contract no longer matches `ComplianceIssue`** (`AnalyzeComplianceText.st:52`) — the prompt asks for `text, category, suggestion`, but there is no `suggestion` field anywhere in the project, and `article` / `explanation` / `action` are no longer requested at all. The compliance popover renders nothing but `article` and `explanation`.

**3. `{userLang}` silently dropped** (`AnalyzeComplianceText.st:9`) — the parameter is still passed from client through resource to service, and the JavaDoc still promises it controls the explanation language. It no longer does anything.

**4. `jobId` has no Bean Validation** (`MapComplianceIssuesRequestDTO.java:13`) — a request without it runs the LLM mapping, returns 200, and persists nothing without any error or log.

A general note on 1–3: the prompt rewrite looks unrelated to "reuse compliance analysis across languages via snippet mapping". Splitting it into its own PR would have surfaced these — the mapping logic itself is sound.

### Test coverage

`62be17d06 "removed tests"` deletes `AiServiceTest.java` (418 lines) including five tests for `mapComplianceIssues`, with no resource-level replacement. That follows the no-`*ServiceTest` rule, but it leaves the empty-source-issues early return, the LLM `catch` branch, the size/null mismatch branch, and — most importantly — the silent dropping of non-verbatim snippets (`continue` after `SnippetMatcher.isVerbatim`) completely uncovered. `SnippetMatcherTest` only covers the trivial helper.

### Things I checked and found fine

Listing these so they don't come up in a later round:

- **LLM call before the ownership check** in `mapComplianceIssues` — consistent with `analyzeJobDescription` and `extractPdfData`, and this PR improves the situation rather than worsening it. `isAdminOrMemberOf` returns `void` and throws `AccessDeniedException`, so nothing is being silently ignored.
- **`let finalContent: string | null = null`** — the type is forced by the pre-existing `extractTranslatedTextFromStream(): string | null`; no lint rule against `null`, and it appears in 77 of 267 client files.
- **The `if (!jobId) return` moved to the top of `translateAndStoreOtherLanguage`** — both callers run after an awaited `saveDraft` with no await in between, so it is unreachable defence, not a behaviour change.
- **`void`-ed analysis promise in `processDescriptionWithAi`** — `analyzeAndUpdateScore` has nothing throwable outside its `try`, and `main` had the same shape with `void Promise.all([...])`.
- **Fully-qualified `java.util.stream.IntStream`** — 19 comparable spots in `src/main/java`, no checkstyle or Spotless gate on imports.
- **`postAndRead(..., Void.class, 400)` without `assertThat`** — `postInvalid` asserts the status itself; 81 test methods in the repo do it this way.
- **`toHaveBeenCalledWith` instead of `toHaveBeenCalledOnce`** — the rule targets call-count assertions; 468 of 546 `toHaveBeenCalledWith` in the repo carry no count assertion.

az108 (commented):
Two follow-up points on the already-resolved review threads — both are convention issues, not functional bugs.

The mapping logic itself still looks right, and the earlier blocking items (`{format}`, the output contract, `{userLang}`, `@NotNull jobId`) all check out as fixed.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1756,8 +1771,9 @@ export class JobCreationFormComponent {
       );
 
       let hasTranslation = false;
+      let finalContent: string | null = null;
```
az108: This line newly introduces `null`, which the client guidelines rule out: *"Use `undefined` exclusively — never use `null`"* (`client-development.mdx:226`).

On `main` this was `const finalContent = ...` scoped inside the `if (accumulatedContent)` block — hoisting it out is what added the explicit `| null` annotation and the `= null` initializer.

The helper's `string | null` return only forces `null` if you store the raw value. Narrowing at the single assignment removes it entirely and simplifies two more expressions:

```ts
let finalContent = '';
...
finalContent = this.extractTranslatedTextFromStream(accumulatedContent) ?? '';
...
if (finalContent.length > 0) {
...
translatedText: extractTextFromHtml(finalContent),
```

Semantics are identical: `null → '' → length === 0` hits the same false branch. Line 1823 already sits behind `hasTranslation === true`, so its `?? ''` was a dead fallback anyway.

src/main/java/de/tum/cit/aet/ai/dto/MapComplianceIssuesRequestDTO.java
```diff
@@ -0,0 +1,16 @@
+package de.tum.cit.aet.ai.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import de.tum.cit.aet.ai.domain.ComplianceIssue;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import java.util.UUID;
+
+@JsonInclude
```
az108: `@JsonInclude` without a value defaults to `Include.ALWAYS`, which is Jackson's built-in default — so this annotation is a no-op and doesn't actually omit anything. It's also the only valueless `@JsonInclude` in the repo; the other 124 all carry an explicit value.

Please use the documented default:

```java
@JsonInclude(JsonInclude.Include.NON_EMPTY)
```

Per `server-development.mdx:104`, `NON_EMPTY` is preferred; `NON_NULL` is only for when empty collections or strings are meaningful and should be sent. Matches `AssignSlotRequestDTO` and `BookSlotRequestDTO`.

az108 (commented):
Image

Its not 2 seconds as you describe it in the pr description but after you fix the two small code issues we can merge this PR

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
az108: Correcting myself here: I listed this exact line under "Things I checked and found fine" in my earlier review. That call was wrong, on two counts.

1. I said the `null` was "forced by the pre-existing `extractTranslatedTextFromStream(): string | null`". That only holds for the *union member*, and only if the raw return value is stored as-is. The `= null` **initializer** was never forced — an initializer is needed here purely for definite-assignment analysis, and `''` satisfies that just as well.

2. My "77 of 267 client files" figure was a bad count. It matched the substring `null` anywhere, so it swept up `toBeNull`, `nullish`, `non-null` and comments. Counting actual `| null` type unions and `= null` assignments in non-spec, non-generated files gives 27 — and most of those are unavoidable browser/Angular API interop (`localStorage.getItem`, `querySelector`, `JSON.parse`, `RegExp.exec`, `DynamicDialogRef`), not a precedent for new code.

Sorry for the churn — that one's on me, not on the PR.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1758,8 +1773,9 @@ export class JobCreationFormComponent {
       );
 
       let hasTranslation = false;
+      let finalContent: string | null = '';
```
az108: we still have null usage here, please adjust

src/main/java/de/tum/cit/aet/ai/dto/MapComplianceIssuesRequestDTO.java
ge94zec: NON_EMPTY does not work here though because an empty complianceIssues list means “clear all issues for this language.” Omitting it makes the field null, causing @NotNull to return 400. The corresponding AiResourceTest reproduces this behavior.
This is the exception described in server-development.mdx: empty collections are meaningful and must be sent. I therefore switched to @JsonInclude(JsonInclude.Include.NON_NULL). I can also remove the annotation entirely if preferred for request DTOs.

PR #2499, opened 2026-05-10
`Bugfix`: Make filter dropdowns close and scroll properly during keyboard nav

### LINKED ISSUE #1944: Improve Keyboard Navigation for Filter and Sort Bar
Image

Should we expand tab navigation? Currently I can navigate with tab to the next filter dropdown and open it. Expected behaviour would be in this case that the previous filter dropdown closes. 

Additionally, we could add auto scrolling to keyboard navigation since currently hidden elements at the bottom are not scrolled into the view. 

_Originally posted by @Bofan-Zhu in https://github.com/ls1intum/tum-apply/pull/1929#discussion_r2829989997_

### LINKED ISSUE #2237: Fix Keyboard Navigation and Selection Logic in Multiselect Components
### Description

The dropdown component currently has broken keyboard accessibility. While the first selection using the Enter key works as expected, attempting to select additional items results in the wrong element being selected. Furthermore, users are currently unable to deselect items using the keyboard, making the component non-functional for those relying on assistive technologies or keyboard-only navigation.

This issue prevents the application from being fully accessible and creates a frustrating, unpredictable experience. Users should be able to toggle selections (select and deselect) and navigate through the list without the focus state becoming desynchronized from the actual selection.

### Tasks

_No response_

### Acceptance Criteria

- [ ] Selecting multiple items in sequence via keyboard targets the correct elements without index shifting.
- [ ] Keyboard users can deselect an active item using the same key trigger used for selection.

az108:
#### General

#### Client

### Motivation and Context

Closes #1944. Closes #2237.

Two related keyboard-navigation defects in `FilterMultiselect`:

- **#1944** — Tabbing from one filter dropdown to the next leaves the previous one open, so multiple dropdown panels overlap. Arrow-key navigation doesn't scroll options into view, so items below the visible area can't be reached without the mouse.
- **#2237** — Pressing Enter on a keyboard-focused option toggles the *wrong* item, because the keyboard handler indexed into `filteredOptions()` (raw filter order) while the template rendered `sortedOptions()` (selected-first then alphabetical). Every toggle re-sorted the list, shifting items under the user.

### Description

All changes are localized to `src/main/webapp/app/shared/components/atoms/filter-multiselect/`:

- **Frozen render order on open.** A new private `renderedOptions` signal snapshots the sorted list when the dropdown opens. The template iterates a single flat `visibleOptions()` (snapshot + live search filter + live `selected` from `selectedValues()`). Toggling updates checkbox state but never reorders. The snapshot is rebuilt on the next open.
- **Single flat option list in the template.** The two `@for` blocks (selected / unselected) collapse into one. The "Selected (N) · Clear all" header still appears above the list when something is selected, but no longer introduces a section break. The keyboard-focus highlight binds on the unified loop with the iteration index, which now matches `focusedIndexOptionList()`.
- **Keyboard handler indexes the rendered list.** `onTriggerKeydown` reads from `visibleOptions()`, so Enter/Space toggles the visually highlighted item. Pressing Enter on a selected item deselects it.
- **Search-term change resets focus.** `onSearchChange` resets `focusedIndexOptionList` to `-1` so it doesn't point past the new filtered list.
- **Close on focus-out.** A new `(focusout)` host listener closes the dropdown when `relatedTarget` is `null` or outside the component's subtree. Mirrors the existing `(document:click)` outside-click handler. `closeDropdown()` is idempotent, so the two listeners coexist safely.
- **Scroll focused option into view.** A `viewChildren<ElementRef<HTMLElement>>('optionRow')` query collects the rendered rows; an `effect()` watching `focusedIndexOptionList` calls `scrollIntoView({ block: 'nearest', inline: 'nearest' })` on the focused row. Matches native `<select>` behaviour.

`sortedOptions`, `unselectedOptions`, and `hasUnselectedItems` are removed (no longer used). `selectedOptions` is rebuilt to derive directly from `filterOptions()` + `selectedValues()` so the trigger button's chip rendering still works whether the dropdown is open or closed.

### Steps for Testing

Prerequisites:

1. Open TUMApply and visit the doctorate-positions page (where the Studienrichtung / Standort / Betreuer:in filters appear).

Test plan:

1. Open the first filter dropdown (e.g. Studienrichtung) by clicking its trigger or pressing Enter on it.
2. Press Tab. The next filter's trigger should receive focus and the *first* dropdown should close.
3. Open a filter with many options. Press ArrowDown repeatedly. The focused option should always remain visible — items below the fold should scroll into view as you advance.
4. Open a filter, ArrowDown to highlight the second option, press Enter. Exactly that option should be selected (not a different one).
5. Press Enter again on the same now-selected option. It should deselect.
6. With several items selected, press Enter on each of them via ArrowDown navigation. Each Enter should toggle the item the highlight points at — items must NOT jump to the top of the list mid-session.
7. Type into the search box while the dropdown is open. Filtering should work; the keyboard highlight should reset (no stale highlight pointing past the filtered list).
8. Close and reopen the same dropdown. Selected items should now appear at the top (snapshot refreshed on each open).
9. Switch to German. Labels and option text should still render correctly.

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Tab-out closes the previous dropdown
- [ ] ArrowDown scrolls hidden options into view
- [ ] Enter toggles the highlighted item
- [ ] Enter on a selected item deselects it
- [ ] Search resets the keyboard focus
- [ ] Re-open refreshes the selected-first ordering

### Test Coverage

**Warning:** Some tests failed in the PR check. Coverage may be incomplete. Please check the [workflow logs](https://github.com/ls1intum/tum-apply/actions/runs/28736271817).

#### Client

| filter-multiselect.ts | 81.41% | 284 | 38 | 13.4 |

_Last updated: 2026-07-05 09:43:25 UTC_


src/main/webapp/app/shared/components/atoms/filter-multiselect/filter-multiselect.ts
```diff
@@ -80,43 +91,48 @@ export class FilterMultiselect {
...
-    const filtered = this.filteredOptions();
-    const translateLabels = this.shouldTranslateOptions();
+    const search = this.searchTerm().toLowerCase().trim();
 
-    const opts = filtered.map(option => ({
-      value: option,
-      label: translateLabels ? this.translateService.instant(option) : option,
-      selected: selected.includes(option),
-    }));
+    const withLiveSelection = snapshot.map(opt => ({ ...opt, selected: selected.includes(opt.value) }));
```
Cathy0123456789: Can we please avoid the spread operator

Cathy0123456789 (changes requested):
Was it intended to remove the divider? Not necessary, but just making sure
Before:
Image
After:
Image

Cathy0123456789 (commented):
When navigating at first with the arrows, then the selected option has the blue fill, but when tabbing then, the focused object has the blue border instead. And the arrows only navigate the visibility / scroll while the bluely highlighted option just stays which might be confusing

Image

Would it make sense to instead have the tab go to the next dropdown and the arrows to navigate within the dropdown? Not sure about keyboard navigation guidelines though

PR #2501, opened 2026-05-11
`Development`: Let applicants choose jobs per page

### LINKED ISSUE #840: Pagination - Items per Page
### Description
Extend the pagination component with the option to change the amount of items per page.

### Tasks
- [ ] Allow users to select how many items are displayed per page.
- [ ] Set default number of items to 15.

### Acceptance Criteria
- [ ] User can change the number of jobs per page.
- [ ] Default is 15 jobs and persists across navigation.

az108:
#### General

#### Client

- [ ] I added multiple screenshots/screencasts of my UI changes.

### Motivation and Context

Closes #840. The pagination across the app was fixed at the server-default page size — applicants scanning lots of jobs paid for repeated round-trips and had no way to widen the page. The selector is added to the shared `dynamic-table` so every paginated view inherits the option, and the job overview is migrated onto that shared component.

### Description

- `dynamic-table` gains a `rowsPerPageOptions` input (default `[10, 20, 30, 40, 50]`) that surfaces the PrimeNG items-per-page dropdown next to the paginator.
- Opt-in persistence via the new `storageKey` input: when set, `dynamic-table` hydrates `rows` from `localStorage` on init (emitting `rowsHydrated` so the parent can sync its signal) and writes the new value on every lazy-load change.
- `LocalStorageService` exposes generic `loadPageSize(key, fallback, allowed?)` and `savePageSize(key, size)` helpers. The loader validates the stored value (numeric, positive, optional whitelist) so a tampered key falls back gracefully.
- `job-card-list` previously used a raw `<p-table>` as a paginator-only surface above its card grid. It now consumes `dynamic-table` with `storageKey="jobsPerPage"` and passes the new size to the server on the existing lazy-load path.
- Vitest specs cover hydration, persistence-on-change, validation of out-of-range values, the no-storage-key path, and the job-card-list integration with `rowsHydrated`.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply as an applicant.
2. Open the job overview page.

Steps:

1. Confirm the paginator shows the items-per-page dropdown with `10 / 20 / 30 / 40 / 50`.
2. Switch to 30, navigate to another page, and back — the selection sticks.
3. Reload the page (or open it in a new tab) — the previous selection is still applied.
4. Manually clear `jobsPerPage` from `localStorage` and reload — page size falls back to the default.

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Items-per-page dropdown changes the visible count on job overview.
- [ ] Selection persists across navigation and reloads.
- [ ] Default applies when nothing is stored.

### Screenshots

_To be added._


src/main/webapp/app/job/job-overview/job-card-list/job-card-list.component.ts
```diff
@@ -32,9 +34,11 @@ export class JobCardListComponent {
   jobs = signal<JobCardDTO[]>([]);
   totalRecords = signal<number>(0);
   page = signal<number>(0);
-  pageSize = signal<number>(12);
+  pageSize = signal<number>(10);
```
Cathy0123456789: The number was at 12, because there are 6 cards per row. Changing it to 10 / 15 etc. destroys this layout:

Image

Image

az108: You're right, and thanks for catching it — fixed in a0544971.

The overview now has its own sizes, **6 / 12 / 18 / 24**, and opens on 12 again. I left the shared sizes at 5/10/15/20 for the tables, since a table has no row to fill and multiples of six read oddly there; the size dropdown already takes per-view sizes, which is how the dependency list gets its 25/50/100.

One thing I checked while doing it: the cards aren't a fixed grid, they're `flex-wrap` with `min-w-[16.5rem]` inside a `max-w-[104rem]` container. So six per row is the widest case and it drops to five, four, three, two, one as the window narrows. That means no single number divides evenly at every width — 12 and 24 divide by 1, 2, 3, 4 and 6, while 18 leaves a short last row at four per row. Still much better than 10, and 12 stays the default, so the common case is exact.

This also changed the mobile rule. It used to start on a fixed 5, which the overview no longer offers, so it now starts on **the shortest size a view offers** — 5 for tables, 6 for the cards. A phone shows one card per row either way, so nothing is stranded there.

The merge conflict is resolved too. It was this branch's move to the shared table component against main's removal of the translate pipe from this template; I kept the migration and dropped `TranslateModule`, since main had removed the last `| translate` from the template.

PR #2465, opened 2026-05-03
`Bugfix`: Fix excessive api requests triggering genderdecoder

ge94zec:
#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://ls1intum.github.io/tum-apply/developer/client-guidelines/language-guidelines).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://ls1intum.github.io/tum-apply/developer/general-guidelines/pull-request-guidelines).

#### Server

- [x] **Important**: I implemented the changes with a [very good performance](https://ls1intum.github.io/tum-apply/developer/server-guidelines/database-and-performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://ls1intum.github.io/tum-apply/developer/server-guidelines/server-development).

#### Client

- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding and design guidelines](https://ls1intum.github.io/tum-apply/developer/client-guidelines/client-development).

### Motivation and Context

Closes: https://github.com/ls1intum/tum-apply/issues/2379
### Description

This PR refactors and improves the gender bias analysis and across page reloads and api requests. It introduces a new INCLUSIVE / NON_INCLUSIVE classification and replaces the previous response DTO GenderResponse with BiasedIssues, simplifying data handling and decoupling bias detection from language.

Bias analysis is now persisted directly in the database using @ElementCollection with a Liquibase migration, and the server-side flow has been streamlined by refactoring services, improving how language-specific results are merged and stored. Additional repository changes prevent Hibernate reload issues.

On client-side, duplicate/numerous API requests during job creation are reduced through better deduplication and improved execution flow. The job creation flow now analyzes the genderDecoder and after that the source language while running translation and secondary analysis in the background.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply as Prof 
2. Nav to Create Position
3. Verify wit dev tool network that biasedIssues are included in jobId and are persistent across page reloads or switches 

#### Code Review

- [x] Code Review 1

#### Manual Tests

- [ ] Test 1

#### Client

[...]

PR #2520, opened 2026-05-14
`Development`: Add action buttons in popover for recommendation system

### LINKED ISSUE #2348: Add action buttons in popover for recommendation system
### Description

- Shorten and extend analysis prompt with suggestions
-  Add UI 

### Tasks

_No response_

### Acceptance Criteria

_No response_

ge94zec:
#### General

#### Server

#### Client

### Steps for Testing

Prerequisites:

1. Log in to TUMApply as Prof 
2. Nav to Create Position 
3. Verify the action buttons behave correctly and text is saved across page switches

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Test 1

#### Client

| job-creation-form.component.ts | 66.03% | 1260 | 72 | 5.7 |
| editor.component.ts | 74.44% | 387 | 24 | 6.2 |
| ai-compliance-popover.component.ts | 47.61% | 38 | ? | ? |
| suggestion-system.component.ts | 30.00% | 17 | ? | ? |

#### Server

| ComplianceIssue.java | 0.00% | 27 |
| ComplianceScoreService.java | 86.67% | 93 |

_Last updated: 2026-06-10 13:16:51 UTC_


src/main/webapp/app/shared/components/molecules/ai-compliance-popover/ai-compliance-popover.component.html
```diff
@@ -1,10 +1,18 @@
 @if (issue()) {
   <div
-    class="fixed z-50 w-72 rounded-md bg-background-default shadow-lg p-3 text-sm pointer-events-none text-xs"
+    class="fixed z-[1001] w-72"
```
az108: do not use so high z levels

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -885,6 +892,68 @@ export class JobCreationFormComponent {
     this.popoverY.set(event.y);
   }
 
+  onPopoverHovered(isHovered: boolean): void {
+    this.isCompliancePopoverHovered.set(isHovered);
+    if (!isHovered) {
+      this.closeCompliancePopover();
+    }
+  }
```
az108: is this a method beeing used in the html directly without a signal in between?

src/main/webapp/content/scss/prime-ng-overrides.scss
az108, line 1: do you really need all of this styling in the scss as override?

src/main/webapp/app/shared/components/molecules/ai-compliance-popover/ai-compliance-popover.component.ts
az108, line 1: some javadoc style comments would be good here

src/main/java/de/tum/cit/aet/ai/service/ComplianceScoreService.java
```diff
@@ -39,16 +39,32 @@ protected int calculateLegalScore(List<ComplianceIssue> compliance) {
...
+            .filter(i -> ComplianceCategory.DSGVO_MINIMIZATION == i.getCategory())
+            .count();
+
         if (criticalCount > 0) {
             return 0;
         }
 
+        if (dsgvoCount > 0) {
+            return 0;
+        }
```
Cathy0123456789: Can this be not combined to 1 if clause?

src/main/java/de/tum/cit/aet/ai/service/ComplianceScoreService.java
```diff
@@ -39,16 +39,32 @@ protected int calculateLegalScore(List<ComplianceIssue> compliance) {
             .filter(i -> ComplianceCategory.CRITICAL_AGG == i.getCategory())
             .count();
 
+        long dsgvoCount = compliance
+            .stream()
+            .filter(i -> ComplianceCategory.DSGVO_MINIMIZATION == i.getCategory())
+            .count();
```
Cathy0123456789: actually ideally just directly to 1 stream as well, since it seems like you are not using the number further

src/main/java/de/tum/cit/aet/ai/service/ComplianceScoreService.java
```diff
@@ -39,16 +39,32 @@ protected int calculateLegalScore(List<ComplianceIssue> compliance) {
...
         long transparencyCount = compliance
             .stream()
             .filter(i -> ComplianceCategory.TRANSPARENCY == i.getCategory())
             .count();
 
-        double score = 100.0 * Math.pow(PENALTY_FACTOR, transparencyCount);
+        long publicSectorCount = compliance
+            .stream()
+            .filter(i -> ComplianceCategory.PUBLIC_SECTOR == i.getCategory())
+            .count();
```
Cathy0123456789: same here

src/main/webapp/app/shared/components/molecules/ai-compliance-popover/ai-compliance-popover.component.html
Cathy0123456789: 1001? Do we need such a high number?

src/main/webapp/i18n/de/job.json
```diff
@@ -233,6 +233,13 @@
           "warning": "Deine Ausschreibung könnte inklusiver sein. Stellenanzeigen mit höherem Score sprechen vielfältigere Bewerbende an und erhalten mehr Rückmeldungen.",
           "good": "Das ist schon eine gute Grundlage. Deine Ausschreibung ist bereits inklusiv und spricht viele Bewerbende an. Mit gezielten Verfeinerungen kannst du sie noch stärker machen.",
           "excellent": "Eine wirklich gelungene und einladende Ausschreibung! Deine Ausschreibung ist sehr einladend formuliert und spricht eine breite Vielfalt an Bewerbenden an."
+        },
+        "popover": {
+          "dismissButton": "Verwerfe",
+          "replaceButton": "Ersetze",
+          "addButton": "Hinzufügen",
+          "removeButton": "Lösche",
+          "applyButton": "Anwenden"
```
Cathy0123456789: Maybe all buttons should have the same naming with "n" in the end

src/main/webapp/content/scss/prime-ng-overrides.scss
Cathy0123456789, line 1: Do we need to add the stylings in this file? Can they not go there where we need them?

Cathy0123456789 (changes requested):
Thank you, small comments on the code

az108 (changes requested):

az108 (commented):
All ten earlier threads are resolved, and with #2465 and #2470 now merged the diff is finally a reviewable 28 files. Four things, one of them a real behaviour bug — details inline.

### Two 0-byte files are committed

```
src/main/java/de/tum/cit/aet/ai/service/ComplianceScoreService.java      0 bytes
src/test/java/de/tum/cit/aet/ai/service/ComplianceScoreServiceTest.java  0 bytes
```

Both were moved to `ai/util/ComplianceScoreCalculator` by #2465 and deleted on main. Merge commit `936070dc4` hit a modify/delete conflict and resolved it by keeping the files and emptying them. `ComplianceScoreCalculator.java` and its test exist intact alongside, so no coverage is lost — this is pure dead weight.

It doesn't break anything (an empty `.java` is a valid compilation unit, and `compileJava`/`compileTestJava` pass), and `grep` finds zero references to `ComplianceScoreService` anywhere in the repo. They can just be deleted. I couldn't leave an inline comment since there are no lines to anchor to.

### Unrelated, but worth knowing

PR #2638 (`feat/2637-port-new-gender-contracts`) also claims `00000000000053`, with a file that is a byte-identical duplicate of the already-merged 052 and isn't registered in `master.xml` at all — so it would never run. No git conflict with this PR (different filenames, and #2638 doesn't touch `master.xml`), and the defect is on that side. Flagging it here only because it surfaced while checking the numbering.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -948,8 +964,12 @@ export class JobCreationFormComponent {
    * @param lang The current language of the editor content
    */
   private applyHighlights(compliance: ComplianceIssue[] | undefined, lang: string): void {
+    const dismissedIssues = this.dismissedComplianceHighlights();
     const highlights = (compliance ?? []).flatMap(issue =>
-      hasText(issue.text) && issue.category !== undefined && (!hasText(issue.language) || issue.language === lang)
+      !dismissedIssues.includes(issue) &&
```
az108: `dismissedComplianceHighlights` stores object *references* and this filters with `.includes(issue)` — SameValueZero, so reference equality on a generated plain interface with no identity field. That only holds while the array keeps the same references, and the early-return branch added in this PR at line 1847 breaks exactly that: `translateAndStoreOtherLanguage` now calls `mapIssuesToTargetLanguage` even when the text is *unchanged*, and line 1998 replaces the target-language issues with fresh objects from the `mapComplianceIssues` response.

Reachable flow, with no re-analysis anywhere in it:

1. Draft analyzed in EN, translated to DE, DE issues mapped
2. On the DE tab, dismiss a highlight — disappears correctly
3. Back on EN, edit **any other field** (title, location, workload) so autosave fires
4. `performAnalysis('en')` cache-returns: no HTTP, no `set`
5. The new early-return branch runs `mapComplianceIssues`, and line 1998 swaps the DE issues for fresh objects — invisible, since you are on EN
6. Switch to DE: `languageChangeEffect` → `applyHighlights` → **the dismissed highlight is back**

The dismissed object also stays in the array forever as a dead reference.

The four `complianceIssues.set(...)` call sites separate cleanly: 1467 and 1575 are load paths, 2050 is a real analysis (resetting dismissals there is defensible), but 1998 is pure re-mapping of unchanged text and should not clear anything.

Suggestion: key on something stable rather than identity, e.g. `` `${issue.language ?? ''}|${issue.id ?? issue.text ?? ''}` ``. `AiService.java:561` carries `sourceIssue.getId()` through the mapping round-trip, so IDs survive mapping but not a fresh analysis — which is exactly the granularity you want. The `text` fallback matters because `id` is optional and the local gender-bias path does not always set it.

If reset-on-new-analysis *is* intended, worth doing explicitly at 2050 rather than as a side effect of reference equality — that also stops the array growing unbounded.

Dismiss is not covered by the new specs, so this would not have been caught.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.ts
```diff
@@ -1874,6 +1946,62 @@ export class JobCreationFormComponent {
...
+    }
+
+    const acceptedMappedIssues = mappedIssues.filter(issue => hasText(issue.id) && this.pendingMappedActions.has(issue.id));
+    const appliedMappedIssues: ComplianceIssue[] = [];
+    for (const acceptedIssue of acceptedMappedIssues) {
+      const currentTargetHtml = targetLang === 'en' ? this.jobDescriptionEN() : this.jobDescriptionDE();
+      const updatedTargetHtml = applyComplianceSuggestionToHtml(currentTargetHtml, acceptedIssue);
+      if (updatedTargetHtml === undefined) continue;
+
+      this.pendingMappedActions.delete(acceptedIssue.id!);
```
az108:
```ts
this.pendingMappedActions.delete(acceptedIssue.id!);
```

`client-development.mdx:230`: *"Never use the TypeScript non-null assertion operator (`!`) to bypass type checks. It hides potential runtime errors and defeats the purpose of strict typing."*

Worth noting this isn't an established pattern to follow — there is exactly **one** other real `!` assertion in the entire client (`document-cache.service.ts:33`), so this would be the second.

`hasText` *is* a type guard (`text.util.ts:8`), but the `&&` in the filter on line 1974 produces a plain boolean rather than a predicate on `issue`, so `filter` still returns `ComplianceIssue[]` and `id` is back to `string | undefined` in the loop. A local const plus a guard next to the `continue` you already have:

```ts
for (const acceptedIssue of acceptedMappedIssues) {
  const acceptedIssueId = acceptedIssue.id;
  if (!hasText(acceptedIssueId)) continue;
  ...
  this.pendingMappedActions.delete(acceptedIssueId);
```

(ESLint won't flag it — `no-non-null-assertion` is `off` in `eslint.config.mjs:174` — so this is convention rather than a lint break.)

src/main/resources/config/liquibase/changelog/00000000000053_add_suggestion_field_to_complianceIssue.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<databaseChangeLog
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+  <changeSet id="052_add_suggestion_field_to_complianceIssue" author="melissa">
```
az108: The changeset id is prefixed `052_` but the file is numbered `053`. Per `liquibase-guidelines.mdx:41`, the id should start with the same changelog number as the file:

```xml
<changeSet id="053_add_suggestion_field_to_compliance_issue" author="melissa">
```

No functional collision — Liquibase keys on id+author+filename, and there are zero duplicate ids repo-wide — but the convention is actively followed: 29 of the 31 files in the 023–053 range conform.

Two smaller things in the same file while you're in there:

- The filename description is camelCase (`complianceIssue`). It's the only camelCase description among all 54 changelogs; `liquibase-guidelines.mdx:20` asks for snake_case.
- Missing `<preConditions onFail="MARK_RAN">` — I checked and am **not** raising this as a real issue: only 7 of 54 changelogs use them, and a plain `addColumn` on a column that never existed runs exactly once via DATABASECHANGELOG anyway. Mentioning only so it doesn't get raised later.

Confirmed the migration itself applies: `JobResourceTest` runs the full changelog against Testcontainers MySQL, 37 tests pass.

src/main/java/de/tum/cit/aet/ai/service/AiService.java
```diff
@@ -560,6 +565,7 @@ public List<ComplianceIssue> mapComplianceIssues(MapComplianceIssuesRequestDTO r
                     sourceIssue.getArticle(),
                     sourceIssue.getExplanation(),
                     sourceIssue.getAction(),
+                    mappedTexts.get(i * 2 + 1).trim(),
                     request.toLang()
```
az108, line 573: can you add a little bit of java doc explaining the i and the calculations here

src/main/webapp/app/shared/components/atoms/editor/editor.component.ts
```diff
@@ -400,6 +401,47 @@ export class EditorComponent extends BaseInputDirective<string> {
...
+  /**
+   * Applies an AI compliance suggestion to the current editor content based on compliance action.
+   * - `Replace`: swaps the target snippet with the suggestion
+   * - `Remove`: deletes the target snippet
+   * - `Add`: inserts the suggestion after the target snippet
+   *
+   * @param issue The compliance issue containing the action, snippet text, and suggestion.
+   * @returns The updated editor HTML for issued text snippet
+   */
+  public applyComplianceSuggestion(issue: ComplianceIssue): string | undefined {
```
az108: Design point: this puts page-specific domain logic into a shared atom, and the coupling here is a different kind from what the editor carried before.

`EditorComponent` lives in `shared/components/atoms/` and is consumed by eight features — research-group template edit, research-group info, research-group detail view, research-group creation form, review dialog, interviewee assessment, application creation page 3, and job creation. Only the last one has anything to do with compliance.

On `main` the editor imported exactly two compliance symbols:

```ts
ComplianceIssueDTOCategoryEnum        // pick the highlight underline colour
ComplianceIssueDTOCategoryEnumValues
```

That is narrow and defensible — the atom needs *a* category the same way it needs a font size. This PR adds `ComplianceIssue` (the whole DTO) and `ComplianceIssueActionEnum`, and `applyComplianceSuggestion` switches on `issue.action` to decide what REPLACE / REMOVE / ADD *mean*. The atom went from "knows compliance categories for styling" to "knows the compliance issue model and executes its business rules".

**The PR already contains the better design.** `compliance-suggestion.util.ts` implements `applyComplianceSuggestionToHtml(html, issue)` — the same rule (locate snippet, switch on action, Replace/Remove/Add with sentence-end insertion) as a pure function. This method even imports `findSentenceEnd` from it, so the extraction was started and then stopped halfway. The result is one rule with two implementations and two different effectors: the Quill API here, DOM `Range` there.

**And they already disagree**, which is the part that is a bug today rather than a preference. When the snippet is not found:

| | ADD action, snippet not found |
| --- | --- |
| `editor.component.ts:435` | `insertAt = editor.getLength()` — appends the sentence at the **end of the document** |
| `compliance-suggestion.util.ts:28` | `if (start === -1) return undefined` — **no change at all** |

`onComplianceSuggestionAccepted` runs both paths for a single accept: this method for the active language, the util for the paired-language HTML. So accepting one ADD suggestion whose snippet is missing in the other language bolts a sentence onto the end of the active editor and silently does nothing to the other. That falls straight out of having the rule in two places.

Suggested shape — keep the atom's real job ("apply a text edit to the live Quill document") and drop the domain:

```ts
// atom, domain-free
public applyTextEdit(edit: { index: number; deleteLength: number; insert: string }): string | undefined
```

The page then computes the edit from the issue via the util, which already knows how. That removes `ComplianceIssue` and `ComplianceIssueActionEnum` from the atom entirely, collapses the two implementations into one, and makes the EN/DE paths symmetric — which fixes the divergence above as a side effect rather than leaving it to surface later.

For context, the gender-decoder surface already on this component (`showGenderDecoderButton`, `biasedAnalysis`, `genderDecoderClick`, `showAnalysisModal`) is the same pattern and predates this PR, so this is a direction worth stopping rather than something introduced here from scratch.

az108 (changes requested):

az108 (approved):

PR #2521, opened 2026-05-14
`Bugfix`: Fix resend code action after cooldown in otp registration modal

### LINKED ISSUE #1383: Fix resend code action after cooldown in otp registration modal
### Description

In the email verification modal of the registration flow (Step 2 of 4), a cooldown is immediately applied for Resend code. After the cooldown timer finishes, clicking on Resend code incorrectly advances the user to Step 3 in an unauthorised state instead of simply resending a new OTP and keeping the user on Step 2.

Image

Image

Expected behaviour
Resend code should only trigger a new OTP email and remain on Step 2 regardless of cooldown. The user should only move to Step 3 after successfully submitting a valid OTP.

Current behaviour
After waiting for the cooldown to end, clicking Resend code navigates to Step 3 while the user is not yet verified or authorised. This breaks the verification flow and causes confusion.

### Tasks

1. Investigate the click handler for Resend code in the OTP modal.
2. Confirm whether the handler incorrectly reuses the logic that advances to Step 3 or calls a shared flow action.
3. Ensure that only a successful OTP verification response triggers navigation to Step 3.

### Acceptance Criteria

- [ ] After the cooldown finishes, clicking Resend code sends a new OTP email and keeps the user on Step 2.
- [ ] The user never navigates to Step 3 as a result of clicking Resend code alone.
- [ ] Step 3 is only reachable when the OTP verification API confirms a valid code and the user is authorised.

SultanTheHL:
#### General

#### Client

### Motivation and Context
Closes https://github.com/ls1intum/tum-apply/issues/1383

### Description

The registration flow's OTP sender always incremented the LoginStep, and didn't check if it was a resend. I decided to add a boolean flag, which when true, skips the nextStep call.
If the resend flow is different enough, we could also decouple it from the initial requestOtp flow. 

### Steps for Testing

1. Try to register on TumApply
2. Enter your email
3. Wait 60 seconds and click on resend code

#### Client

| auth-facade.service.ts | 3.05% | 276 | 58 | 21.0 |
| otp-input.ts | 100.00% | 103 | ? | ? |

_Last updated: 2026-05-24 09:21:24 UTC_


az108 (changes requested):
The styling of the modal is broken and tumapply up is shown in white and therefore not visible, could you fix that aswell please 

Image

src/main/webapp/app/shared/components/organisms/login/login.html
```diff
@@ -1,4 +1,4 @@
-<div class="flex items-center justify-center self-stretch p-4 text-[2rem] font-bold rounded-t-xl bg-primary text-text-on-primary">
+<div class="flex items-center justify-center self-stretch p-4 text-[2rem] font-bold rounded-t-xl bg-primary text-white">
```
az108: please do not use colors directly, as tumapply supports various themes, we have semanticc tokens which you can have a look at and use

src/main/webapp/app/shared/components/organisms/registration/registration.html
```diff
@@ -1,4 +1,4 @@
-<div class="flex items-center justify-center self-stretch p-4 text-[2rem] font-bold rounded-t-xl bg-accent text-[var(--p-text-on-accent)]">
+<div class="registration-modal-header flex items-center justify-center self-stretch p-4 text-[2rem] font-bold rounded-t-xl text-white">
```
az108: same here

src/main/webapp/app/shared/components/organisms/registration/registration.scss
```diff
@@ -1,3 +1,7 @@
+.registration-modal-header {
+  background-color: var(--p-accent-color);
+}
+
```
az108: we want to avoid scss as much as possible, if its possible via tailwind then we prefer that please, and this should be possible as there should be a semantic token equivalent for this background color

az108 (changes requested):

az108 (changes requested):
Now the bug with proceeding to step 3 when resend code is pressed is gone, thank you for that. But when i press resend code after a 60 seconds wait timer then i can spam the button and multiple emails can be sent without a timer. Can we please make sure that when we resend the the code that the 60 seconds timer is triggered again. And can we please reduce the timer to 30 seconds.

Image

src/main/webapp/content/scss/prime-ng-overrides.scss
```diff
@@ -16,22 +16,27 @@ p-step[data-p-active='true'] .p-step-title {
...
+  @include auth-modal-icon-color(var(--p-text-on-primary));
 
-.login-header-button .p-button-text.p-button-info:hover,
-.registration-header-button .p-button-text:hover {
-  filter: drop-shadow(0 0 0.075rem white);
-  background: rgba(255, 255, 255, 0.05);
+  .p-button:hover,
+  button.p-button:hover {
+    filter: drop-shadow(0 0 0.075rem var(--p-text-on-primary));
+    background: rgba(255, 255, 255, 0.08);
```
az108, line 38: and while your at it you could replace this with a semantic token white which we have. Then we get rid of the hardcoded white in our codebase

az108 (changes requested):

az108 (approved):
issue mentioned was fixed in another PR 👍

PR #2529, opened 2026-05-14
`Development`: Use real buttons and links instead of click-handler divs

### LINKED ISSUE #2395: Remove duplicated click and keyboard handlers for interactive containers
### Description

Several templates implement the same accessibility pattern manually on non-semantic interactive elements such as div, p, or article: adding role, tabindex, click handling, and keyboard activation for Enter and sometimes Space.
This duplication increases maintenance cost and makes behavior inconsistent across the app. Some elements behave like buttons, some like links, and some need custom event handling. We should define a shared, reusable pattern for keyboard-activatable non-native elements and apply it only where semantic HTML cannot be used directly.
The goal is to improve consistency and accessibility while avoiding behavior regressions. The solution should not blindly prevent default or stop propagation for all interactions, and it should preserve cases that require custom roles or custom event handling.

### Tasks

1.  Audit current uses of clickable non-semantic elements with manual role, tabindex, click, and keyboard bindings
2. Group cases by behavior: button-like, link-like, and custom/special-case interactions.
3. Decide the standard approach for each group
4. Use native <button> or <a> where feasible.
5.  Introduce a shared directive or utility only for non-semantic elements that must remain non-semantic.

### Acceptance Criteria

- [ ] Repeated manual keyboard-activation patterns for suitable non-semantic interactive elements are reduced through a shared approach.

az108:
#### General

#### Client

### Motivation and Context

Closes #2395.

Across the app many templates implemented the same accessibility pattern by hand on non-semantic elements (`<div>`, `<p>`, `<article>`): `role="button"` + `tabindex="0"` + `(click)` + `(keydown.enter)`, sometimes with an extra `(keydown.space)`. That duplication makes behaviour inconsistent (e.g. one row also handled Space, the next didn't; `<a>` tags without `href` had to fake Enter activation themselves) and offers a worse experience than a real `<button>` or `<a>` (no right-click "open in new tab" for links, no native keyboard contract, no automatic focus ring on every browser).

### Description

The audit found 23 occurrences across 16 files. They fell into three buckets and were handled accordingly:

1. **Button-like** (10) — converted to `<button type="button">` with a small new `.btn-bare` utility class in `global.scss` that strips the default button chrome so the existing card/row styling is preserved. Affected: `sidebar-button`, `image-upload-button`, `upload-button` (outer + the redundant nested `role=button` span), `rating` segments, `string-input` helper-text-right link, `selectable-slot-card`, `interview-process-card`, `document-dialog` list rows, `profile-picture-settings` avatar trigger, `research-group-creation-form` user candidate rows, and the `login` / `registration` "switch view" links (these were `<a>` without `href`, so they were never a real link anyway).
2. **Link-like** (4) — converted to `<a [routerLink]>`. Affected: footer `imprint` / `privacy` / `about-us` (the navigation methods on `FooterComponent` were removed since `routerLink` replaces them), and `job-card` (the `onViewDetails` / `onKeyDown` pair was replaced by a `[routerLink]="detailLink()"`).
3. **Custom / non-semantic by necessity** (6) — kept as `<div>` but now use a new shared `ClickableDirective` (`[jhiClickable]`) that owns `role`/`tabindex` host bindings and the Enter/Space activation. These are the cases where the element has an interactive descendant (a checkbox, a `<jhi-button>`, an `<a>`), which the HTML spec forbids inside `<button>` and `<a>`. Affected: `interviewee-card` (inner action buttons), `interviewee-section` and `assign-applicant-modal` and `research-group-add-members` rows (inner checkbox), `filter-multiselect` (trigger with chips + option rows with checkbox), `upcoming-interview-card` (inner anchor), and two locations in `job-creation-form` (image card and AI compliance banner).

Three dead `stopPropagation`-only wrappers around delete-button overlays (`research-group-images`, `department-images` x2) were just removed — their parent cards aren't even clickable, so the wrapper did nothing.

The directive emulates native button keyboard semantics (Enter always, Space only when `role="button"`) and uses `event.preventDefault()` only on the keys it actually handles, so it does not blindly suppress other interactions — matching the issue's explicit guidance.

The `@angular-eslint/template/click-events-have-key-events` rule was extended with `ignoreWithDirectives: ['jhiClickable']` so the directive is a first-class way to satisfy the rule without inline keyboard handlers.

### Steps for Testing

Prerequisites:

1. Log in to TUMApply.
2. Visit each affected page and confirm both mouse and keyboard activation still work:
   - Sidebar items, footer links, login/register switch links, file/image upload buttons, rating segments, string-input "helper-text-right" link.
   - Job overview cards (clicking opens the detail; right-click → "Open in new tab" now works), job-creation image picker (selecting + deleting), document-dialog item list, profile picture editor.
   - Interview process cards, upcoming-interview cards, selectable slot cards, interviewee row + interviewee card (still navigates to assessment; inner buttons still work without selecting the card).
   - Filter multiselect (open/close, select/deselect options).
   - Research group "Add members" row selection, research group creation form admin candidate rows.
3. Tab through each surface — every previously clickable element is still reachable and activates with Enter (Space too for buttons).

#### Code Review

- [ ] Code Review 1

#### Manual Tests

- [ ] Mouse activation for every converted element behaves identically to before.
- [ ] Keyboard activation works (Tab to focus, Enter to activate; Space activates buttons but not links).
- [ ] No visual regressions on cards, rows, or icon buttons.

#### Client

| interviewee-card.component.ts | 0.00% | 61 | 11 | 18.0 |
| interviewee-section.component.ts | 0.00% | 360 | 41 | 11.4 |
| assign-applicant-modal.component.ts | 0.00% | 142 | ? | ? |
| upcoming-interview-card.component.ts | 0.00% | 43 | 9 | 20.9 |
| job-creation-form.component.ts | 0.00% | 1251 | 88 | 7.0 |
| job-card.component.ts | 100.00% | 78 | 8 | 10.3 |
| footer.component.ts | 0.00% | 27 | 9 | 33.3 |
| filter-multiselect.ts | 0.00% | 219 | 30 | 13.7 |
| clickable.directive.ts | 100.00% | 27 | 8 | 29.6 |
| research-group-add-members.component.ts | 0.00% | 179 | 40 | 22.3 |
| sidebar-button.component.ts | 95.00% | 36 | 2 | 5.6 |
| jobs-preview-section.component.ts | 100.00% | 51 | 12 | 23.5 |

_Last updated: 2026-07-13 19:57:45 UTC_


src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
Cathy0123456789, line 460: I guess it is related to this part of the code, but with tabbing I cannot delete an image from the Image Collection

src/main/webapp/app/job/job-overview/job-card/job-card.component.html
Cathy0123456789, line 1: Not sure if relevant, but earlier, you could navigate with Space as well, now that is not possible anymore

src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
```diff
@@ -537,7 +530,7 @@ <h1 jhiTranslate="{{ pageTitle() }}"></h1>
           [shouldTranslate]="true"
         />
       } @else if (aiToggleSignal() && complianceCount() > 0) {
-        <div tabindex="0" (click)="handleMessageClick($event)" (keydown.enter)="handleMessageClick($event)">
+        <div jhiClickable tabindex="0" (click)="handleMessageClick($event)">
```
Cathy0123456789: This does not work anymore as it seems

src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
az108: Good catch, and it was worse than just the delete not firing — pressing Enter on the trash icon **selected the image** instead.

Keyboard events bubble, and `jhiClickable` listened for Enter/Space on the card. So a key press on the nested delete button ran the card's handler, which called `preventDefault()` (cancelling the button's own native activation) and then clicked the card. The directive exists precisely for wrappers that have to contain a button or link, so it had the bug in the one case it was written for.

Fixed in the directive rather than here: it now ignores key presses that did not happen on the host itself, so nested buttons and links keep handling their own keys. Added tests for it in `clickable.directive.spec.ts`.

src/main/webapp/app/job/job-overview/job-card/job-card.component.html
az108: Relevant, but I'd argue it's now correct and I've left it as is — happy to change it if you disagree.

The card is a real `<a [routerLink]>` now. Native anchors activate on **Enter** only; Space scrolls the page, and that's what keyboard and screen-reader users expect from a link. The old Space behaviour came from it being a `role="button"` div with a manual keydown handler, which is the pattern this PR removes.

Re-adding Space would mean putting a key handler back on an anchor and swallowing page scroll. The honest alternative would be making the card a `<button>` that navigates programmatically, but that loses middle-click, open-in-new-tab and the status-bar URL preview — so a link seems like the right trade.

src/main/webapp/app/job/job-creation-form/job-creation-form.component.html
az108: Confirmed, and thanks — this was a real regression.

The link inside the translated message is `<a tabindex="0" class="... stepper-link">` with no `href`, so it never fires a click of its own. The old code caught the **bubbled** keydown on the wrapper and recognised the link via `event.target`'s class. Routing it through `jhiClickable` instead synthesised a click on the wrapper, so `event.target` became the wrapper and the class check could never match.

(I checked that Angular's sanitizer keeps both the `tabindex` and the class on that anchor, so the link really is the tab stop.)

The wrapper is a delegator, not a control, so it's back to plain click + Enter delegation with `role="presentation"`. Dropping `jhiClickable` also removes the second, dead tab stop it had added around the whole banner.

src/main/webapp/app/shared/components/molecules/document-dialog/document-dialog.html
```diff
@@ -22,21 +22,18 @@
             />
             -->
             <!-- Label container -->
-            <div
-              class="min-w-0 grow cursor-pointer px-3 py-2 rounded-md hover:text-text-on-primary hover:bg-primary-default"
+            <button
+              type="button"
+              class="btn-bare min-w-0 grow cursor-pointer px-3 py-2 rounded-md hover:text-text-on-primary hover:bg-primary-default w-full text-left"
```
Cathy0123456789: You seemed to have added w-full and text-left, do we need it?

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
```diff
@@ -154,25 +154,24 @@
 
             @if (showCandidatesList()) {
               @for (user of adminProfessorCandidates(); track user.id) {
-                <div
-                  class="flex items-center gap-3 p-3 cursor-pointer transition-colors"
+                <button
+                  type="button"
+                  class="btn-bare w-full text-left flex items-center gap-3 p-3 cursor-pointer transition-colors"
```
Cathy0123456789: Here also w-full and text-left?

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
```diff
@@ -154,25 +154,24 @@
 
             @if (showCandidatesList()) {
               @for (user of adminProfessorCandidates(); track user.id) {
```
Cathy0123456789: For a long list of professors, the user would have to tab through all of them to be able to continue to the next fields as it seems. Maybe it would make more sense to scroll in the list with arrow keys and use tab to go to the next fields?

src/main/webapp/app/shared/components/molecules/document-dialog/document-dialog.html
az108: Both dropped in 1bb2ff9c — you're right, neither was doing anything.

This button sits in `<li class="flex min-w-0 items-center gap-3 ...">` and already carries `grow`, so it fills the remaining width as a flex item; `w-full` was a second, redundant way of saying it. And `btn-bare` sets `text-align: inherit`, which is what undoes the browser's `text-align: center` on buttons, so `text-left` was repeating that too.

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
az108: Same here, both dropped.

Slightly different reason for `w-full` though: this list is `<div class="flex flex-col ...">`, and flex items stretch across the cross axis by default, so in a column they already fill the width. `text-left` is redundant for the same reason as the other one — `btn-bare` restores the inherited alignment.

Worth a quick look when you next pull, since I reasoned this from the flex context rather than measuring it in a browser.

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
az108: Agreed that this is a problem, and it's the one item from your review I've deliberately not changed yet — it's a bigger call than the others and I'd rather agree the shape first.

You're describing the listbox pattern: the list becomes a single tab stop, arrows move between options, Enter or Space selects, and Tab leaves for the next field. That is the right pattern for a single-select list like this, and it's what the APG prescribes.

Two things make me want to check with you before doing it:

1. **It should cover the add-members dialog too.** That has the same shape of list, and having one candidate list answer to arrows while the other answers to Tab would be worse than either choice applied consistently.
2. **Arrow-key lists are only discoverable if they look like lists.** Once the list stops being a series of tab stops, the selected option needs a clear focus style, otherwise a keyboard reader lands on the container and cannot tell that arrows do anything.

So: happy to implement it across both lists, with a visible focused-option style. Say the word and I'll do it here, or I can split it into its own PR if you'd rather not grow this one — it's a behaviour change rather than the semantic-markup swap the rest of this PR is.

az108: Done for both lists in 7420c08b.

Each list is now a single tab stop: arrows move between people, Home and End jump to the ends, Enter or Space picks one, Tab carries on to the next field. The lists are `role="listbox"` and the rows `role="option"` with `aria-selected`, so the selection is announced rather than left to the highlight — and both draw a focus outline, since an option can now be reached without the pointer.

Two things worth knowing beyond the mechanical change:

**The add-members list had the worse version of this problem.** Every row was two tab stops, not one — the row itself plus the checkbox inside it. So a page of 5 people cost 10 stops. The row is the control now and the checkbox only shows state, so it's out of the tab order and off the accessibility tree (it needed a small `tabIndex` input on `jhi-checkbox` to get there). That list is `aria-multiselectable`, since it picks several.

**The tabbable option is clamped rather than reset.** If you're on the eighth result and search again for something with three, a stored index of 8 would leave no option tabbable and the list would become unreachable by keyboard. It clamps to the last option instead, so the list always has exactly one way in.

The key handling is a small pure function (`nextOptionIndex`) shared by both lists with 13 tests, including that recovery case. Movement stops at the ends rather than wrapping, so you can feel where the list ends.

Worth a keyboard pass when you next pull — particularly whether the focus outline reads clearly against the selected-row tint.

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
```diff
@@ -153,27 +153,35 @@
...
-                  <div
-                    class="flex-shrink-0 w-8 h-8 rounded-full bg-primary-default flex items-center justify-center text-text-on-primary font-semibold"
+              <!-- One tab stop for the whole list; the arrow keys move between candidates. -->
+              <div role="listbox" [attr.aria-label]="candidateListLabel()" class="flex flex-col">
+                @for (user of adminProfessorCandidates(); track user.id; let index = $index) {
+                  <button
+                    #candidateOption
+                    type="button"
+                    role="option"
+                    class="btn-bare flex items-center gap-3 p-3 cursor-pointer transition-colors focus-visible:outline-2 focus-visible:-outline-offset-2 focus-visible:outline-primary"
```
Cathy0123456789: Using outline-primary is I think a bit inconsistent, since we don't do it anywhere else, but it's okay

src/main/webapp/app/usermanagement/research-group/research-group-add-members/research-group-add-members.component.html
```diff
@@ -52,29 +52,32 @@
...
-            <div class="flex-1">
-              <div class="font-medium text-text-primary">{{ user.displayName }}</div>
-              <div class="text-sm text-text-secondary">{{ user.email }}</div>
+        <!-- One tab stop for the whole list; the arrow keys move between people. -->
+        <div role="listbox" aria-multiselectable="true" [attr.aria-label]="userListLabel()">
+          @for (user of users(); track user.id; let index = $index) {
+            <div
+              #userOption
+              role="option"
+              class="flex items-center gap-3 p-3 cursor-pointer transition-colors focus-visible:outline-2 focus-visible:-outline-offset-2 focus-visible:outline-primary"
```
Cathy0123456789: same here

src/main/webapp/content/scss/global.scss
```diff
@@ -9,6 +9,22 @@
...
 
+/* For native <button> elements used purely as clickable containers (cards, list rows,
+   icon triggers). Only covers what Tailwind's preflight leaves behind: buttons come out
+   centred and, since v4, with a default cursor. Background, border, padding, margin, font
+   and colour are already reset there, so repeating them here only creates something that
+   can fight the utilities on the same element.
+
+   Kept in the components layer for that same reason. Unlayered rules outrank every layered
+   one whatever their specificity, so declaring this outside a layer would let it beat the
+   utilities instead. Components sits before utilities, so utilities win. */
```
Cathy0123456789: Maybe a bit long comment

src/main/webapp/app/shared/components/molecules/research-group-creation-form/research-group-creation-form.component.html
az108: Fair point, and it was slightly worse than "we don't do it anywhere else" — we *do* have a focus style elsewhere, just a different idiom. `info-icon.component.ts` on main uses `focus-visible:ring-2 focus-visible:ring-primary-default focus-visible:ring-offset-1`. So this was a second, competing idiom, pasted into two templates as a bare utility string.

Pulled it into a named class in the components layer in 25cc8190, next to `.btn-bare`:

```scss
.focus-ring-inset:focus-visible {
  outline: 2px solid var(--p-primary-color);
  outline-offset: -2px;
}
```

The compiled CSS is unchanged: `outline-primary` resolves to `--color-primary-DEFAULT` → `var(--p-primary-color)`, which is the same colour `ring-primary-default` already uses. So it's a pure refactor — same pixels, just no longer copy-pasted, and now a primitive rather than a one-off.

One thing I did keep deliberately: the **inset** offset. These rows sit flush against each other inside a scrolling container, so an outset ring like info-icon's `ring-offset-1` would be clipped at the container edge and overlap the neighbouring row. That's why this one draws inside, and why I named it `-inset` rather than folding it into a generic `.focus-ring`.

I did not convert info-icon over: outset is the right choice for a standalone icon with space around it, and it's outside this PR's diff. Happy to unify them into one primitive with two variants in a follow-up if you'd prefer.

src/main/webapp/app/usermanagement/research-group/research-group-add-members/research-group-add-members.component.html
az108: Same change here — this row now carries the shared `.focus-ring-inset` class instead of the three inline utilities (25cc8190).

Reasoning is in the [thread on the creation form](https://github.com/ls1intum/doc-apply/pull/2529#discussion_r3668687255): same compiled output, and the inset offset is deliberate because these rows sit flush inside a scrolling container.

src/main/webapp/content/scss/global.scss
az108: Agreed — nine lines of prose for a two-line rule. Trimmed in 25cc8190: the `.btn-bare` explanation is down from nine lines to three, and the layering rationale is hoisted to a two-line note above the block, since it now covers the focus-ring rule as well.

It was also partly duplicated. The same "unlayered rules outrank every layered one, so this has to stay in the components layer" argument is already written up in the JSDoc of `btn-bare-layer.spec.ts`, which is the thing actually enforcing it. I left the long version there and kept the SCSS side to a sentence.

